PR #2, opened 2022-02-21
Refactor command line parsing + add trigger support

michaelcfine:
Added support for pre- and post- triggers
As part of doing this, refactored command line argument parsing into its own module

### DIFF
--- a/server/src/commandLine.mjs
+++ b/server/src/commandLine.mjs
@@ -1,0 +1,71 @@
+/*
+* Copyright 2021 Comcast Cable Communications Management, LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


michaelcfine:
👍 but the links on the three buttons/images need to be corrected.

server/src/triggers/lifecycle.close/post.mjs
```diff
@@ -0,0 +1,25 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+* SPDX-License-Identifier: Apache-2.0
+*/
+
+function post(ctx, params) {
+  ctx.setTimeout(function() {
+    const result = { state: 'inactive', previous: ctx.fireboltCoreSdk.Lifecycle.state() };
```
jlacivita, line 21: FYI, the Core SDK manages the value of `previous` so you can leave it out if you want.

jlacivita (dismissed):
LGTM

server/src/triggers/lifecycle.close/post.mjs
michaelcfine: Good call. Have removed 'previous' from both post.mjs files.

server/src/commandLine.mjs
```diff
@@ -0,0 +1,71 @@
+/*
```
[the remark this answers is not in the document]
michaelcfine: This isn't the top-level file you run, though. Okay to leave as-is?

jlacivita (approved):

PR #9, opened 2022-03-23
Add docs about browser extension and triggers

michaelcfine:
Added doc info about triggers
Updated/added info about the browser plugin existing

### DIFF
--- a/docs/Triggers.md
+++ b/docs/Triggers.md
@@ -1,0 +1,100 @@
+Mock Firebolt: Triggers
+=======================
+
+# Overview
+
+You may define pre- and/or post- triggers that fire before and after Firebolt method calls are responded to by Mock Firebolt. These are "hooks" for you to add your own logic and/or actually alter or replace the response that would otherwise be returned by Mock Firebolt.
+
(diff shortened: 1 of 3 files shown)


cletustboone:
Is it intentional that all the content of the `docs/Overview.md` docs is inside of triple backticks? It doesn't appear so since I see markdown links in the content.

docs/Triggers.md
```diff
@@ -0,0 +1,100 @@
...
+Example:
+```
+cd server
+npm run dev -- --triggers ./src/triggers --triggers /some/other/dir/with/triggers
+```
+
+
+# File Organization
+
+Within each directory you pass via `--triggers` command line arguments, you must create subdirectories whose names match Firebolt method names (e.g., `lifecycle.ready`, `device.id`, etc.) and within these subdirectories you must create JavaScript files named `pre.mjs` and/or `post.mjs` depending on which trigger(s) you are defining.
```
cletustboone, line 26: The module names are now capitalized if that makes any difference. So it's `Lifecycle.ready()` now.

michaelcfine: Good to know. I'm going to leave things as they are right now so the docs match with the specific version of the SDK I'm using, but when I go to upgrade to a new version of the SDK, I'll make this change too.

cletustboone (approved):
🚀

docs/Triggers.md
```diff
@@ -0,0 +1,100 @@
...
+# The Context Object
+
+When Mock Firebolt invokes your triggers, it passes your function a context object as the first parameter that includes functions you can use (call) within your triggers.
+
+
+setTimeout
+: The standard setTimeout function; included for your convenience
+
+setInterval
+: The standard setInterval function; included for your convenience
```
cletustboone, line 58: `setInterval` and `setTimeout` functions are already in the global namespace. Why do they need to be on `ctx` for convenience?

PR #11, opened 2022-04-14
Add cors to allow requests from firecert app or similar

michaelcfine:
Use cors package to allow requests from other origins, e.g., from the firecert app or similar.
Issue originally reported and code fix supplied by Neerada Neelakantan.

### DIFF
--- a/server/src/index.mjs
+++ b/server/src/index.mjs
@@ -73,6 +73,7 @@ const __dirname = path.resolve();
 import express from 'express';
 import bodyParser from 'body-parser';
 import { engine } from 'express-handlebars';
+import cors from 'cors';
 
 import * as configureAPI from './configureAPI.mjs';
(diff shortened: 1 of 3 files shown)


server/src/index.mjs
```diff
@@ -83,6 +84,8 @@ app.use(bodyParser.json({ limit: '1mb' })); // to support JSON-encoded bodies
 app.use(bodyParser.urlencoded({             // to support URL-encoded bodies
   extended: true
 }));
+app.use(cors());
+app.options('*',cors());
```
cletustboone, line 88: So allowing all origins by default?

michaelcfine: Yes; seems fine given it's a local/development tool.

cletustboone (approved):

PR #12, opened 2022-04-22
Fix typo in revert and fix state urls

michaelcfine:
Fix typo in revert function
Fix readme and comments to fix state-related urls (add /state/ in urls) to match actual routes

### DIFF
--- a/server/src/routes/api/state.mjs
+++ b/server/src/routes/api/state.mjs
@@ -39,7 +39,7 @@ function getState(req, res) {
   });
 }
 
-// POST /api/v1/global/latency
+// POST /api/v1/state/global/latency
 // Expected body: { latency: { min: xxx, max: xxx } }
 function setLatency(req, res) {
   const userId = getUserIdFromReq(req);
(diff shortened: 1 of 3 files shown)


server/src/routes/api/state.mjs
```diff
@@ -39,7 +39,7 @@ function getState(req, res) {
   });
 }
 
-// POST /api/v1/global/latency
+// POST /api/v1/state/global/latency
```
cletustboone, line 42: Do you have to do anything to tell your server about this new piece of the route's path? How does it know that the handlers for `api/state/` are located here? Is that just by convention of the dir structure?

cletustboone (approved):
Looks okay on this end.

server/src/routes/api/state.mjs
michaelcfine: There's a "configureAPI.mjs" file that has the actual Express routes. Here, I'm just updating the comments to match the actual routes.

PR #17, opened 2022-05-06
Use ^0.6.1 for core and use github links for manage and discovery

michaelcfine:
- Moved from 0.6.0 to 0.6.1 for core
- Added "^" to pin to major version (0), meaning no more package.json updates till v1 (hopefully)
- Changed manage and discovery links to github links since we're not putting artifacts for those SDKs into npm for now


jlacivita (approved):
LGTM

PR #20, opened 2022-05-16
Feature session recording

TrikkStar:


michaelcfine (changes requested):
Looks great with the exception that I think we should move the session-related functions out of handleMessage.mjs and into its own sessionManagement.mjs file (and then you can either merge your sessionOBJ code into that same, single file or keep them separate if you really want).

michaelcfine (approved):
👍 Looks great. Seems cleaner with the session management stuff all in one file. Nice!

PR #23, opened 2022-05-20
Use portable scripts in package.json, not shell commands

michaelcfine:
Since Windows users were/are having trouble with the shell script commands used in the package.json scripts, this PR changes these to use portable NodeJS utility scripts.

If anyone has temporarily removed the "2>/dev/null || :" bit in the script for build:mf, they should revert their change to avoid a merge conflict and should be able to use the new package.json file as-is on Windows.


TrikkStar (approved):
Everything runs fine in basic PowerShell now.

PR #24, opened 2022-05-24
Added support for mixed case module name

sourabhk2399:
Support for mixed case module names in OpenRPC specification files and in JSON RPC messages at runtime


michaelcfine (approved):
👍 Looks good!

PR #27, opened 2022-06-09
Added support for sequence of events

sourabhk2399:
Added support for sequence of events


cli/src/cli.mjs
```diff
@@ -312,7 +312,46 @@ if ( parsed.help ) {
...
+    const sEvent = fs.readFileSync(path.resolve(__dirname, eventFile), {encoding:'utf8', flag:'r'});
+    let seqevent;
+    if ( eventFile.endsWith('yaml') || eventFile.endsWith('yml') ) {
+      seqevent = yaml.load(sEvent);
+    } else {
+      seqevent = JSON.parse(sEvent);
+    }
+    msg(`Sending sequence of events based on file ${eventFile}...`);
+    //iterating through sequence of events
+    for(let i=0;i<seqevent.length;i++) {
```
michaelcfine, line 327: The way things are now, you can only use the CLI to submit a sequence of events... You loop here and do <n> posts to the /sequence endpoint. But this means that someone can't create a single post request with a body like what's inside a sequence file and I think we always want to support the rest api first and then treat the CLI as "syntactic sugar" (as much as possible). So... can you instead (a) here, do a single post of the file body as-is, (b) move this loop logic down into code that runs regardless of whether the user uses a rest post call or the cli (your sequenceManagement.mjs file, I think) and obviously instead of looping and posting, loop and call sendEvent.

michaelcfine (changes requested):
See comment. Mainly: Let's make sure a single post to /sequence with a body that is the same as what's in the event file(s) works. Then treat the CLI as a simple read-the-file-put-the-contents-into-a-post-body-and-do-a-post usage of that core rest feature.

sourabhk2399:
Hi Mike,
I have corrected the changes as per your suggestion.
Please verify the same (There are 2 commits for it, please go through both)

Also please send the manage-sdk Zip file.

Thanks and Regards,
Sourabh Khandelwal


From: ***@***.***>
Sent: Wednesday, June 15, 2022 9:04 PM
To: ***@***.***>
Cc: Khandelwal, ***@***.***>; ***@***.***>
Subject: [External] Re: [rdkcentral/mock-firebolt] Added support for sequence of events (PR #27)

This message is from an EXTERNAL SENDER - be CAUTIOUS, particularly with links and attachments.



@michaelcfine requested changes on this pull request.

See comment. Mainly: Let's make sure a single post to /sequence with a body that is the same as what's in the event file(s) works. Then treat the CLI as a simple read-the-file-put-the-contents-into-a-post-body-and-do-a-post usage of that core rest feature.


In cli/src/cli.mjs<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_rdkcentral_mock-2Dfirebolt_pull_27-23discussion-5Fr898132278&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=g46FB7VbscgJbS9Bx5W7Db-0yFkweVPQ5JM3WUVkTr8m_kaWH3XQROdCY8hGagL5&s=tuuJUZn99mPBWMFx7P0xDXEaxzaKITJwbKHQtE4FVu4&e=>:

> @@ -312,7 +312,46 @@ if ( parsed.help ) {

   }



 } else if ( parsed.sequence ) {

-  // @TODO

+  const eventFile = parsed.sequence;

+

+  try {

+    const sEvent = fs.readFileSync(path.resolve(__dirname, eventFile), {encoding:'utf8', flag:'r'});

+    let seqevent;

+    if ( eventFile.endsWith('yaml') || eventFile.endsWith('yml') ) {

+      seqevent = yaml.load(sEvent);

+    } else {

+      seqevent = JSON.parse(sEvent);

+    }

+    msg(`Sending sequence of events based on file ${eventFile}...`);

+    //iterating through sequence of events

+    for(let i=0;i<seqevent.length;i++) {

The way things are now, you can only use the CLI to submit a sequence of events... You loop here and do posts to the /sequence endpoint. But this means that someone can't create a single post request with a body like what's inside a sequence file and I think we always want to support the rest api first and then treat the CLI as "syntactic sugar" (as much as possible). So... can you instead (a) here, do a single post of the file body as-is, (b) move this loop logic down into code that runs regardless of whether the user uses a rest post call or the cli (your sequenceManagement.mjs file, I think) and obviously instead of looping and posting, loop and call sendEvent.

—
Reply to this email directly, view it on GitHub<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_rdkcentral_mock-2Dfirebolt_pull_27-23pullrequestreview-2D1007721503&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=g46FB7VbscgJbS9Bx5W7Db-0yFkweVPQ5JM3WUVkTr8m_kaWH3XQROdCY8hGagL5&s=Pr1HiHhqE1-_iwTXCtQv8-0znrMr2KvFs3H81b1Vecw&e=>, or unsubscribe<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_notifications_unsubscribe-2Dauth_AYTVFIRJL4UJOUB5Y2NGKDTVPHZXJANCNFSM5YJ4NN4Q&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=g46FB7VbscgJbS9Bx5W7Db-0yFkweVPQ5JM3WUVkTr8m_kaWH3XQROdCY8hGagL5&s=uBY9Pni2BN15fwEI2sEjc9O154fClnB3KVyO5LusqF8&e=>.
You are receiving this because you authored the thread.Message ID: ***@***.***>


________________________________

This message is for the designated recipient only and may contain privileged, proprietary, or otherwise confidential information. If you have received it in error, please notify the sender immediately and delete the original. Any other use of the e-mail by you is prohibited. Where allowed by local law, electronic communications with Accenture and its affiliates, including e-mail and instant messaging (including content), may be scanned by our systems for the purposes of information security and assessment of internal compliance with Accenture policy. Your privacy is important to us. Accenture uses your personal data only in compliance with data protection laws. For further information on how Accenture processes your personal data, please see our privacy statement at https://www.accenture.com/us-en/privacy-policy.
______________________________________________________________________________________

www.accenture.com

server/src/configureAPI.mjs
```diff
@@ -77,6 +77,11 @@ function configureAPI(app) {
 
     // End a session
     app.post('/api/v1/session/stop',                   sessionApi.stopSession);
+
+    // ======================= Sequence-Related API Routes =======================
+
+    // Send a sequence
```
michaelcfine, line 83: Minor: Please change this to "Send an event sequence" to be more explicit. Same with line 81... add "Event " before "Sequence".

server/src/routes/api/event.mjs
```diff
@@ -26,17 +26,19 @@ import * as events from '../../events.mjs';
 
 // POST /api/v1/event
 // Expected body: { method: 'device.onDeviceNameChanged' result: ... }
-function sendEvent(req, res) {
+function sendEvent(req, res, sendSuccessFlag = true) {
```
michaelcfine, line 29: It's usually a bad idea to mix the signature of an Express route ( f(req, res) ) with any other use of parameters. Ultimately, we want this to be a regular route handler signature (no extra param) and it will probably just call sequenceManagement.sendSequence once it gets the req.body value.

server/src/routes/api/event.mjs
```diff
@@ -26,17 +26,19 @@ import * as events from '../../events.mjs';
...
   const { ws } = res.locals; // Like magic!
   const { method, result } = req.body;
 
   function fSuccess() {
```
michaelcfine, line 33: All of the fSuccess, etc. stuff only exists in the event case because we had the weird case where we are reusing the events.sendEvent function between here and the ctx.sendEvent helper function. We don't need that stuff here. This function should basically just be the HTTP-related stuff (get body, parse params if any (no), etc.; eventually return 200 or 400 or whatever). But in the middle it should just call a function like sequence.sendSequence. Nice and simple.

michaelcfine (changes requested):
Still some work to do. Ultimately, what we want is:
- the cli to read the file and call sequenceManagement.mjs:: executeSequence
- the REST API to read the HTTP POST body and call sequenceManagement.mjs::executeSequence
- the sequenceManagement.mjs::executeSequence function to (a) take the data structure as authored by the user (with a mix of 'at' and 'delay'), (b) do the preprocessing to compute 'at' values for all events, (c) loop through the array of events and send them (using events.mjs:: sendEvent

Only the route (routes/api/sequence.mjs) should deal with "HTTPisms" (req, res; req.body, HTTP response codes, etc.).

cli/src/usage.mjs
```diff
@@ -38,7 +38,7 @@ const lines = [
   { cmdInfo: "--method device.id --errCode -32888 --errMsg \"Sad day for you\"", comment: "Set error code & msg for method" },
   { cmdInfo: "--upload ../examples/slow.json  ",                                 comment: "See examples/ directory for, uh, examples" },
   { cmdInfo: "--event ../examples/device-onDeviceNameChanged1.event.json",       comment: "Send event (method, result keys expected)" },
-  { cmdInfo: "--sequence ../examples/events1.sequence.json  ",                   comment: "See examples/events1.json" },
+  { cmdInfo: "--sequence ../examples/events1.sequence.json  ",                   comment: "Send an event sequence (See examples/events1.json)" },
```
michaelcfine, line 41: Is there a real sequence example file name to use here?

server/README.md
```diff
@@ -395,3 +395,42 @@ curl --location --request POST 'http://localhost:3333/api/v1/event' \
...
+
+
+## Send an event sequence
+
+### Example cURL Command:
+
+```
+curl --location --request POST 'http://localhost:3333/api/v1/sequence' \
+--header 'Content-Type: application/json' \
+--data-raw '[{
```
michaelcfine, line 407: Just a question, really: Do I really post "just" the array, or do I post an object with some well-known key (seqevent?) whose value is the array?

server/src/routes/api/sequence.mjs
```diff
@@ -0,0 +1,31 @@
...
+import {executeSequence} from '../../sequenceManagement.mjs';
+
+//Execute sequence events with respective delay values
+function sendSequence(req, res) {
+    let seqevent = req.body
+    const { ws } = res.locals; // Like magic!
+    const userId = getUserIdFromReq(req);
+
+    //iterating through sequence of events
+    for(let i = 0; i < seqevent.length; i++) {
```
michaelcfine, line 12: I'm still confused by the presence of this loop here. I would expect this code to basically just pass the incoming data structure (in req.body) to the underlying "shared code" in sequenceManagement.mjs.

server/src/sequenceManagement.mjs
```diff
@@ -0,0 +1,11 @@
+import {sendEvent} from './events.mjs';
+
+function executeSequence(ws,userId,method_name, result, msg, atTime_val) {
+
+  setTimeout(function() {
```
michaelcfine, line 5: I would expect the loop here.

michaelcfine (changes requested):
Getting closer, but it still feels like the loop is in the route rather than in the underlying 'library' function. I think it's better that the sequenceManagement.mjs file is the only one that really 'understands' the datastructure. The CLI just calls the REST endpoint and the REST endpoint (after reading the body) just calls the executeSequence function in sequenceManagement.mjs and then that's where the magic happens.

sourabhk2399:
Hi Mike,
I have pushed the changes, Please review it.

Thanks and Regards,
Sourabh Khandelwal
From: ***@***.***>
Sent: Thursday, June 23, 2022 5:07 AM
To: ***@***.***>
Cc: Khandelwal, ***@***.***>; ***@***.***>
Subject: [External] Re: [rdkcentral/mock-firebolt] Added support for sequence of events (PR #27)

This message is from an EXTERNAL SENDER - be CAUTIOUS, particularly with links and attachments.



@michaelcfine requested changes on this pull request.

Getting closer, but it still feels like the loop is in the route rather than in the underlying 'library' function. I think it's better that the sequenceManagement.mjs file is the only one that really 'understands' the datastructure. The CLI just calls the REST endpoint and the REST endpoint (after reading the body) just calls the executeSequence function in sequenceManagement.mjs and then that's where the magic happens.


In cli/src/usage.mjs<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_rdkcentral_mock-2Dfirebolt_pull_27-23discussion-5Fr904384378&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=a96aM0B-yEI7YleNc1W_U3lB5l4oXO2nTpROybKfC2A&e=>:

> @@ -38,7 +38,7 @@ const lines = [

   { cmdInfo: "--method device.id --errCode -32888 --errMsg \"Sad day for you\"", comment: "Set error code & msg for method" },

   { cmdInfo: "--upload ../examples/slow.json  ",                                 comment: "See examples/ directory for, uh, examples" },

   { cmdInfo: "--event ../examples/device-onDeviceNameChanged1.event.json",       comment: "Send event (method, result keys expected)" },

-  { cmdInfo: "--sequence ../examples/events1.sequence.json  ",                   comment: "See examples/events1.json" },

+  { cmdInfo: "--sequence ../examples/events1.sequence.json  ",                   comment: "Send an event sequence (See examples/events1.json)" },

Is there a real sequence example file name to use here?


In server/README.md<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_rdkcentral_mock-2Dfirebolt_pull_27-23discussion-5Fr904385582&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=dlZkySF-b25D_jo1OTvCCH5PxXvgg0d3rM3Gev54Jnc&e=>:

> @@ -395,3 +395,42 @@ curl --location --request POST 'http://localhost:3333/api/v1/event'<https://urldefense.proofpoint.com/v2/url?u=http-3A__localhost-3A3333_api_v1_event-27&d=DwQCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=FWMRuhkUl9AfpFvMmJ5T_3VIu-YsY54p7q-NbEStYc4&e=> \

     "status": "OK"

 }

 ```

+

+

+## Send an event sequence

+

+### Example cURL Command:

+

+```

+curl --location --request POST 'http://localhost:3333/api/v1/sequence'<https://urldefense.proofpoint.com/v2/url?u=http-3A__localhost-3A3333_api_v1_sequence-27&d=DwQCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=eKW5sbJxproUulhdqIwQYyfGysOfUIrL6-QycaUdEpU&e=> \

+--header 'Content-Type: application/json' \

+--data-raw '[{

Just a question, really: Do I really post "just" the array, or do I post an object with some well-known key (seqevent?) whose value is the array?


In server/src/routes/api/sequence.mjs<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_rdkcentral_mock-2Dfirebolt_pull_27-23discussion-5Fr904386404&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=uJwq0LI_rNXGQEhm2VmIvKFP5q2jD2_5Pmes2MVQCEE&e=>:

> @@ -0,0 +1,31 @@

+'use strict';

+import { getUserIdFromReq } from '../../util.mjs';

+import {executeSequence} from '../../sequenceManagement.mjs';

+

+//Execute sequence events with respective delay values

+function sendSequence(req, res) {

+    let seqevent = req.body

+    const { ws } = res.locals; // Like magic!

+    const userId = getUserIdFromReq(req);

+

+    //iterating through sequence of events

+    for(let i = 0; i < seqevent.length; i++) {

I'm still confused by the presence of this loop here. I would expect this code to basically just pass the incoming data structure (in req.body) to the underlying "shared code" in sequenceManagement.mjs.


In server/src/sequenceManagement.mjs<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_rdkcentral_mock-2Dfirebolt_pull_27-23discussion-5Fr904387186&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=zeCJdgpuXUI6p1y0rPnojhMQ6_m7MXgDJvCx4UDj8kE&e=>:

> @@ -0,0 +1,11 @@

+import {sendEvent} from './events.mjs';

+

+function executeSequence(ws,userId,method_name, result, msg, atTime_val) {

+

+  setTimeout(function() {

I would expect the loop here.

—
Reply to this email directly, view it on GitHub<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_rdkcentral_mock-2Dfirebolt_pull_27-23pullrequestreview-2D1016114986&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=4am3f3xb4aFxFL43tDe_lKlX4tAi2u_7KDCwuKcDIs8&e=>, or unsubscribe<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_notifications_unsubscribe-2Dauth_AYTVFIVYDPIRWAU42KD2H3LVQOPSVANCNFSM5YJ4NN4Q&d=DwMCaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=vAd8F272GG1U8SxPUesbPGWZacX7s3okUBKYhy6v7TFcj6sgE80-MEGMr7hyyE2d&m=-v9azSNipivyYzUsC_4Ct9FxfbhqsFVSQJ_dyorVgadeP4FzxHnH9g12V50n_vFV&s=X2-brpjIMdfgKY5UtYh4LcTGia_GAitUuSerJD1RX90&e=>.
You are receiving this because you authored the thread.Message ID: ***@***.***>


________________________________

This message is for the designated recipient only and may contain privileged, proprietary, or otherwise confidential information. If you have received it in error, please notify the sender immediately and delete the original. Any other use of the e-mail by you is prohibited. Where allowed by local law, electronic communications with Accenture and its affiliates, including e-mail and instant messaging (including content), may be scanned by our systems for the purposes of information security and assessment of internal compliance with Accenture policy. Your privacy is important to us. Accenture uses your personal data only in compliance with data protection laws. For further information on how Accenture processes your personal data, please see our privacy statement at https://www.accenture.com/us-en/privacy-policy.
______________________________________________________________________________________

www.accenture.com

server/src/routes/api/sequence.mjs
```diff
@@ -0,0 +1,57 @@
...
+
+import { logger } from '../../logger.mjs';
+import {executeSequence} from '../../sequenceManagement.mjs';
+import * as commonErrors from '../../commonErrors.mjs';
+
+//Execute sequence events with respective delay values
+function sendSequence(req, res) {
+
+    try {
+        executeSequence(req,res);
```
michaelcfine, line 31: We want to keep all code that's NOT in /routes unaware of the req and res objects, which are Node/Express/HTTP/REST objects/concepts. So if you look at the other routes, they generally plug the passed in information (either from the body, url params, or url itself) and then pass these to the "helper function" (that's NOT in routes/).

So the code you have in server/src/sequenceManagement.mjs lines 29 through 37 (rummaging through the req and res objects should be here, and then the API signature for the executeSequence function should take those pieces of info (ws, userId, and seqevent).

server/src/sequenceManagement.mjs
```diff
@@ -0,0 +1,58 @@
...
+
+'use strict';
+
+
+import {sendEvent} from './events.mjs';
+import { getUserIdFromReq } from './util.mjs';
+
+function executeSequence(req,res) {
+  let seqevent
+  const { ws } = res.locals; // Like magic!
```
michaelcfine, line 29: See comment above... this block of code which gets info out of req and res should be in the route function and then this function should just take ws, userId, and seqevent as params (I think; you should confirm these are the params you need/want.)

sourabhk2399: Sure, considered your suggestions and made changes. Please review

server/src/routes/api/event.mjs
```diff
@@ -55,7 +55,7 @@ function sendEvent(req, res) {
     });
   }
 
-  events.sendEvent(ws, userId, method, result, '', fSuccess, fErr, fFatalErr);
+  events.sendEvent(ws, method, result, `${method}`, fSuccess, fErr, fFatalErr);
```
michaelcfine, line 58: It looks like "userId" was dropped somehow. It should be passed, I believe, looking at the signature of the sendEvent function.

michaelcfine (changes requested):
Sooo close. It looks like a parameter got deleted/dropped somehow. That should be the only change left.

server/src/routes/api/event.mjs
sourabhk2399: Changed the code.

PR #29, opened 2022-06-24
OTT-1527

warbhe-amit:


server/src/commandLine.mjs
```diff
@@ -95,8 +98,15 @@ if ( enabledTriggerPaths.length > 0 ) {
   logger.info(`Triggers will be read from these paths: ${enabledTriggerPaths.join(', ')}`);
 }
 
+// --- novalidate method overrides
+let validateMethodOverrides = true;
+if( !config.dotConfig.supportedSdks[0].validateFlag && parsed.novalidate ){
```
michaelcfine, line 103: The validateFlag should be at the top-level of the dotConfig object, not within a supported SDK. So remove ".supportedSdks[0]".

server/src/stateManagement.mjs
```diff
@@ -365,7 +366,10 @@ function validateNewState_MethodOverrides(newStateMethods) {
   for ( const [methodName, methodOverrideObject] of Object.entries(newStateMethods) ) {
     errors = errors.concat(validateMethodOverride(methodName, methodOverrideObject));
   }
-
+// returns an empty array in "novalidate mode"
+  if( !validateMethodOverrides ){
```
michaelcfine, line 370: You need to check both the command-line setting here AND the dotconfig object flag. What you have right now ignores anything in the dotconfig file, I think.

docs/DotConfig.md
```diff
@@ -0,0 +1,48 @@
+Mock Firebolt: mf.config.json(Supported SDK's)
```
michaelcfine, line 1: This file isn't quite what I had in mind, but it's okay for now. Once this PR gets merged (after changes based on this review), I can go in and make a few edits.

server/src/stateManagement.mjs
michaelcfine: Also... you should do this check and possible return at the top of the function; no need to do all the validation only to throw away.

michaelcfine (changes requested):
See comments. The main things are: (1) the dotconfig flag should be at the top level of the dot config object (not in a supported SDK) and (2) the check in state management as to whether to validate should take both the command line flag and the dot config flag into consideration and should be moved up to the start of the function.

docs/Documentation.md
```diff
@@ -56,6 +56,10 @@ For notes about how Firebolt events work, see [Events.md](./Events.md).
 
 Pre- and post- triggers may be defined for any Firebolt method call. See [Triggers.md](./Triggers.md) for more information.
 
+## novalidate Flag
+
+Use this flag for novalidate mode where mock firebolt does not validate uploaded method overrides. This will be used by the certification test suite, not by normal app developer users.
```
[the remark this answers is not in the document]
warbhe-amit: Done

server/src/.mf.config.SAMPLE.json
```diff
@@ -6,7 +6,8 @@
             "cliFlag": null,
             "cliShortFlag": null,
             "fileName": "firebolt-open-rpc.json",
-            "enabled": true
+            "enabled": true,
+            "validateFlag": true
```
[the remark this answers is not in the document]
warbhe-amit: Done

server/src/commandLine.mjs
warbhe-amit: Done

server/src/stateManagement.mjs
warbhe-amit: This check I have taken care in commandLine.mjs:
if( !config.dotConfig.validateFlag && parsed.novalidate ){
  validateMethodOverrides = false;
}
Please confirm.

docs/DotConfig.md
warbhe-amit: okay

cli/src/cli.mjs
```diff
@@ -261,29 +261,35 @@ if ( parsed.help ) {
...
-        console.log(response.data);
-      })
-      .catch(function (error) {
-        logError(error);
-      });
-  } catch ( ex ) {
-    console.log(`ERROR: File ${uploadFile} is either missing or contains invalid JSON or YAML`);
-    console.log(ex);
-  }
+  const uploadDir = parsed.upload  || [];
```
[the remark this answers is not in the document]
warbhe-amit: Done

server/src/stateManagement.mjs
warbhe-amit: Done, Please confirm.

server/src/commandLine.mjs
```diff
@@ -69,6 +72,12 @@ const parsed = nopt(knownOpts, shortHands, process.argv, 2);
 const httpPort = parsed.httpPort || config.app.httpPort;
 const socketPort = parsed.socketPort || config.app.socketPort;
 
+// --- novalidate method overrides
+if(!config.dotConfig.validateFlag || parsed.novalidate ){
```
michaelcfine, line 76: I like this better than what I had in mind, but seeing it in front of my eyes now, I think we should change dotConfig.validateFlag to dotConfig.validateMethodOverrides so the name matches the config setting/property.

michaelcfine (changes requested):
Looking really good. One minor thing, please: See comment. Basically, please rename the dot config validateFlag property name to validateMethodOverrides so it matches the name at the top level of config. So, just be sure to find all instances of 'validateFlag' and rename. Thanks!

server/src/commandLine.mjs
warbhe-amit: Done

PR #30, opened 2022-06-30
Initial unit tests

michaelcfine:
- Created initial "framework" for running unit tests
  - Enabled "npm run test" (package.json change)
  - Initial jest config, etc. (but see notes below)
- Created initial tests for util.mjs and magicDateTime.mjs

- NOTES:
  - I have no idea if the various config files are all necessary, whether we need more or fewer settings in each, etc.
  - For some reason, coverage reports come back as 0% coverage; we should fix this so we can track %


warbhe-amit (approved):
Looks good. One thing: else part is not considered for config.app.defaultUserId for function getUserIdFromReq.

PR #32, opened 2022-07-07
bug fix for invalid event listeners

sourabhk2399:
The issue was in MessageHandler.mjs -> HandleMessage Function. For any event, the check was whether is contains ".on" . Mfos was registering the invalid event (as it contains .on ) and the HandleMessage function ends there .Function was not executing the actual validation for any method/event because this validation. So, I changed the position of code validating the method/event.


michaelcfine (approved):
👍 Looks good to me.

PR #33, opened 2022-07-07
Feature/unit tests

warbhe-amit:
added test cases

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -453,3 +453,24 @@ curl --location --request POST 'http://localhost:3333/api/v1/user'
     "userId": "<uuid>"
 }
 ```
+## Get list of users (returns a list of all the registered users)
+
+### Example cURL Command:
+
+```
+curl --location --request GET 'http://localhost:3333/api/v1/user'
+```
+
+### 200 Response:
+
+```
+{
+    "status": "SUCCESS",
+    "users": [
+        "<uuid1>",
+        "<uuid2>",
+        "<uuid3>",
+        "<uuid4>"
+    ]
+}
+```
--- a/server/src/configureAPI.mjs
+++ b/server/src/configureAPI.mjs
@@ -67,6 +67,10 @@ function configureAPI(app) {
 
     app.post('/api/v1/user',                            userApi.addUser);
 
+    // =======================  User-Related Routes ===========================
+
+    app.get('/api/v1/user',                             userApi.getUsers);
+
 	// ======================= Event-Related API Routes =======================
 
     // Send an event
--- a/server/src/routes/api/user.mjs
+++ b/server/src/routes/api/user.mjs
@@ -44,8 +44,18 @@ function addUser(req, res) {
   });
 }
 
+//GET /api/v1/user
+function getUsers(req, res) {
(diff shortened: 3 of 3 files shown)


server/test/suite/config.test.mjs
```diff
@@ -0,0 +1,44 @@
+"use strict";
+import { config } from "../../src/config.mjs";
+
+test("should first", () => {
```
michaelcfine, line 4: I'll defer to Anushri, but these kinds of tests seem like they're fragile and not worth writing. I think it's okay to leave it, I just don't see the benefit, really.

server/test/suite/events.test.mjs
```diff
@@ -0,0 +1,130 @@
+"use strict";
+
+import { jest } from "@jest/globals";
+import * as events from "../../src/events.mjs";
+import { logger } from "../../src/logger.mjs";
+
+test("events.registerEventListener is working properly", () => {
+  const spy = jest.spyOn(logger, "debug");
+  const dumyObject = {
```
michaelcfine, line 9: "Dummy" has two Ms; if you're in the code making other changes, you can/should change, but it's obviously not high priority.

server/test/suite/fireboltOpenRpc.test.mjs
```diff
@@ -0,0 +1,42 @@
+"use strict";
+import { jest } from "@jest/globals";
+import * as fireboltOpenRpc from "../../src/fireboltOpenRpc.mjs";
+
+test("fireboltOpenRpc.getRawMeta working properly", () => {
+  const expectedResult = {};
+  const result = fireboltOpenRpc.getRawMeta();
+  expect(result).toEqual(expectedResult);
```
michaelcfine, line 8: How does this work? Doesn't seem like it should...

michaelcfine (approved):
Looks good, although I didn't examine super closely. Let's get code coverage working so we can see where we're at.

server/test/suite/events.test.mjs
warbhe-amit: done

server/test/suite/fireboltOpenRpc.test.mjs
warbhe-amit: In this fireboltOpenRpc.mjs file, getRawMeta function is returning an object which can't be accessible from test files as its blank object initially, that's why its returning an empty object and we are just checking with deep equality.
Please confirm.

server/test/suite/fireboltOpenRpc.test.mjs
```diff
@@ -0,0 +1,735 @@
...
+    "invalidMethod",
+  ];
+  const expectedOutput = [true, true, false];
+  expectedInput.forEach((methodName, index) => {
+    const result = fireboltOpenRpc.isMethodKnown(methodName);
+    expect(result).toBe(expectedOutput[index]);
+  });
+});
+
+// test('fireboltOpenRpc.getSchema', () => {
```
warbhe-amit, line 292: getSchema_error

@michaelcfine, getting this error while running the test case.

PR #35, opened 2022-07-08
Add support for user groups

michaelcfine:
- Invented new userId 'syntax'... "<userId>~<groupName>" (notice that's a tilde, not a dash)
- Method calls from one (full) userId should cause method responses only to that client/app using that userId
- Events heading to one (full) userId should go to all userIds that have the same embedded groupName value
- In events.mjs::sendEvent, instead of sending one message just for the given ws, look up all the ws's in the same group and send messages to each

- Added some doc to docs/UsageWithinApps.md
- Added new doc page, docs/UserGroups.md

- Added a map from groupName to list of full userIds in userManagement.mjs to make it easy to find all full userIds in the same group as the group of a given full userId
- Because the apps may send method calls and event listeners in any order, the message IDs might be different... so, changed the eventListenerMap data structure in events.mjs to be a map of maps, with the top-level map being full userId and the inner maps being what was there originally... a map from event name to message id

- In index.mjs, added "123~A", "456-A", "789-A" so the QA team can use these out-of-the-box (e.g., use 123~A for the certification app and 456~A for the app under test).


michaelcfine:
To test, use any app (e.g., https://github.comcast.com/ottx/simple-firebolt-test-app suitably edited to have some event listener code which console.logs when it is called) and then run like this:

http://localhost:8081?mf=ws://localhost:9998/123-A
http://localhost:8082?mf=ws://localhost:9998/456-A

Note that the "core" user id values (123, 456) are different, which keeps method call mock overrides separate, but they are in the same "user group" (A).

PR #37, opened 2022-07-12
provide support to use proxy

vyvanand:
**commandLine.mjs**
- "proxy" mode support added for CLI "--proxy <ip>"
**index.mjs**
- If Proxy mode, Get token from connection parameter or env. If not found, process the request as is. 

**messageHandler.mjs**
1. Handle incoming message which will by pass default OPENRPC calls in case of proxy enabled. 
2. Initialize websocket connection to proxy server and handle failure
3. Send inbound messages to proxy endpoint.
4. Send outbound message from proxy endpoint to caller.
 
**server/src/proxyManagement.mjs**
1. Establish proxy connection which would forward request to websocket server
2. If connection not established, reject the event. 
3. If connection established, Listen for incoming messages and send request to proxy endpoint.
4. In case of any disconnect with connection, reinitalize ws connection after 2s and process inbound request same as above. 

**server/src/config.mjs** - Update on this novalidate mode.
novalidate mode where Mock Firebolt skips certain validation steps, based on options you provide.

When Mock Firebolt starts up,it will check for `validate` value in mf.config.json file and novalidate flag passed via the `--novalidate` command line argument.

validate: [ "method", "params", "response", "events" ].

`validate` array string in the mf.config.json file will take precedence over the --novalidate command-line argument, if given.


server/src/index.mjs
```diff
@@ -24,23 +24,48 @@ import './developerTools.mjs';
...
+    userId = config.app.defaultUserId;
+    // If connection parameter doesn't have token, then SSH and retrieve
+    const token = await proxyManagement.getThunderToken(request)
+    if( token ) {
+      if( token.stdout ) {
+        process.env.wsToken = token.stdout
+      } else {
+        logger.error(`ERROR: Unable to get thunder token: ${token.stderr}`);
+        socket.destroy();
+        process.exit(-1)
```
kschrief, line 64: What's the reason for the negative exit code?

server/src/messageHandler.mjs
```diff
@@ -47,6 +48,31 @@ async function handleMessage(message, userId, ws) {
     return;
   }
 
+  //bypass JSON-RPC calls and directly target device. 
+  if(process.env.proxy) {
+    let params = oMsg.params
+    const payload = {
```
kschrief, line 54: Is there a reason for re-creating the message and not passing it on as-is?

server/src/messageHandler.mjs
```diff
@@ -47,6 +48,31 @@ async function handleMessage(message, userId, ws) {
...
+      id: oMsg.id
+    }
+    if(params) {
+      payload.params = params
+    }
+    
+    let wsProxy = await proxyManagement.getProxyWSConnection()
+    if( ! wsProxy ) {
+      //init websocket connection for proxy request to be sent and update receiver client to send request back to caller.
+      wsProxy = await proxyManagement.initialize(ws)
```
kschrief, line 66: What if initialization fails? We should catch the error and fail gracefully here.

server/src/util.mjs
```diff
@@ -45,6 +45,13 @@ function createTmpFile(prefix, postfix) {
   return tmpObj;
 }
 
+function validateIPaddress(ipaddress) {
```
kschrief, line 48: Does this regex cover IPv6?

server/src/proxyManagement.mjs
```diff
@@ -0,0 +1,191 @@
...
+import child_process from 'child_process';
+const exec = promisify.promisify(child_process.exec);
+import { parse } from 'url';
+import WebSocket from 'ws';
+let websocketConnection = null 
+
+async function initialize(receiverWSClient) {
+    const url = await buildWSUrl()
+    if (url) {
+      console.log("Establishing websocket connection")
```
kschrief, line 31: Can this be a little more descriptive considering we now have several connections to deal with? "Downstream" or "Proxy" connections come to mind.

server/src/proxyManagement.mjs
```diff
@@ -0,0 +1,191 @@
...
+        ws.addEventListener('open', openCallback)
+
+        ws.onopen = function () {
+            console.log("connected to websocket client") 
+            websocketConnection = ws
+        }
+
+        ws.onclose = function(){
+            // connection closed, discard old websocket and create a new one in 5s
+            console.log("connection closed")
```
kschrief, line 59: Can this be a bit more descriptive?

server/src/proxyManagement.mjs
```diff
@@ -0,0 +1,191 @@
...
+        ws.on('message', function message(data) {
+            var buf = Buffer.from(data);
+            const oResponseMessage = {
+              jsonrpc: '2.0',
+              id: JSON.parse(buf.toString()).id,
+              ...JSON.parse(buf.toString())
+            };
+            const responseMessage = JSON.stringify(oResponseMessage);
+            receiverWSClient.send(responseMessage);
+            console.log(`Sent message to ripple: ${responseMessage} \n`);
```
kschrief, line 80: Can this be a bit more generic? Something like "Sent outbound message"

server/src/proxyManagement.mjs
```diff
@@ -0,0 +1,191 @@
...
+        stderr: '',
+      };
+      // If token already exists, return token
+      if(process.env.wsToken) {
+        output.stdout = process.env.wsToken
+        resolve(output)
+      }
+      const { query } = parse(request.url);
+      if(query && query.includes("token=") && query.length > 6) {
+        resolve(true)
```
kschrief, line 164: This is not resolving with a token the way the other "resolve()" calls are. Is this on purpose?

kschrief (changes requested):

server/src/proxyManagement.mjs
vyvanand: Addressed

server/src/proxyManagement.mjs
vyvanand: Addressed

server/src/proxyManagement.mjs
vyvanand: Addressed

server/src/proxyManagement.mjs
vyvanand: Addressed

server/src/util.mjs
vyvanand: Added the change to cover ipv6.

server/src/messageHandler.mjs
vyvanand: Addressed

server/src/messageHandler.mjs
vyvanand: Addressed. Using as is

server/src/index.mjs
vyvanand: It should be 1 for general error. Updated

server/src/index.mjs
```diff
@@ -24,23 +24,48 @@ import './developerTools.mjs';
...
+    }
+    process.env.thunderIP = commandLine.proxy
+    logger.info('Using proxy server ' + process.env.thunderIP);
+    //close websocket connection if already exists
+    proxyManagement.close()
+    process.env.proxy = true
+    userId = config.app.defaultUserId;
+    // If connection parameter doesn't have token, then SSH and retrieve
+    const token = await proxyManagement.getThunderToken(request)
+    if( token ) {
```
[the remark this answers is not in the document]
vyvanand: addressed

server/src/index.mjs
```diff
@@ -24,23 +24,48 @@ import './developerTools.mjs';
...
 import { createServer } from 'http';
 import { parse } from 'url';
-import WebSocket, { WebSocketServer } from 'ws';
 
 logger.important(`Welcome to Mock Firebolt`);
 
 const server = createServer();
 
-server.on('upgrade', function upgrade(request, socket, head) {
+server.on('upgrade', async function upgrade(request, socket, head) {
```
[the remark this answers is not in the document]
vyvanand: to use await, we need this to be async.

kschrief (approved):

kschrief (approved):

michaelcfine (changes requested):
Looks good but a few changes... let's set up some time to chat maybe?

server/src/index.mjs
```diff
@@ -24,23 +24,42 @@ import './developerTools.mjs';
...
+
+  if(commandLine.proxy) {
+    if(!validateIPaddress(commandLine.proxy)) {
+      logger.error('ERROR: Invalid IP address');
+      socket.destroy();
+    }
+    process.env.proxyServerIP = commandLine.proxy
+    logger.info('Send proxy request to websocket server: ' + process.env.proxyServerIP);
+    process.env.proxy = true
+    userId = config.app.defaultUserId;
```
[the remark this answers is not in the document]
vyvanand: Addressed

server/src/messageHandler.mjs
```diff
@@ -47,6 +48,31 @@ async function handleMessage(message, userId, ws) {
     return;
   }
 
+  //bypass JSON-RPC calls and directly target device.
```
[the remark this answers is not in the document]
vyvanand: We do have ticket in this sprint to support method triggers. This ticket is just beginning to provide support for proxy.

README.md
```diff
@@ -158,6 +158,10 @@ npm run dev -- --manage --discovery
 # If you need to use non-standard ports for any reason:
 npm run dev -- --httpPort 3456 --socketPort 9876
 
+# If you need to use proxy connection for any reason:
+npm run dev -- --proxy ip
+Note: To use token run "export TOKEN=<token>" command before running MFOS
```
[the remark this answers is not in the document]
vyvanand: Adressed

README.md
```diff
@@ -158,6 +158,10 @@ npm run dev -- --manage --discovery
 # If you need to use non-standard ports for any reason:
 npm run dev -- --httpPort 3456 --socketPort 9876
 
+# If you need to use proxy connection for any reason:
+npm run dev -- --proxy ip
```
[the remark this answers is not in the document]
vyvanand: Addressed

server/src/events.mjs
```diff
@@ -243,6 +243,11 @@ function coreSendEvent(isBroadcast, ws, userId, method, result, msg, fSuccess, f
 
       const finalResult = ( postResult ? postResult : result );
 
+      // Error to be logged in "novalidate mode"
+      if( ! config.validate.includes("events") ) {
```
michaelcfine, line 247: This code seems wrong. The if part looks almost right (I think you don't want the bang), but it shouldn't automatically call fErr... it should actually validate finalResult and then only call fErr if that result didn't validate, then call fErr. Right?

server/src/fireboltOpenRpc.mjs
```diff
@@ -121,6 +125,10 @@ function getDeveloperNotesForMethod(methodName) {
 // Are the given params valid for thegiven method, based on the OpenRPC metadata?
 // Returns an array of errors; Returns an empty array if no errors are found
 function validateMethodCall(methodName, params) {
+  // Returns an empty array in "novalidate mode"
+  if( ! config.validate.includes("params") ){
```
michaelcfine, line 129: Same as above... I think no "!".

server/src/index.mjs
```diff
@@ -41,13 +41,28 @@ const server = createServer();
...
     userId = config.app.defaultUserId;
   }
+  
+  if( commandLine.proxy ) {
+    process.env.proxyServerIP = commandLine.proxy
+    logger.info('Send proxy request to websocket server: ' + process.env.proxyServerIP);
+    process.env.proxy = true
+    // Get token from connection parameter or from env
+    const token = proxyManagement.getToken(request)
+    if( token.stdout ) {
```
michaelcfine, line 59: This seems like a weird protocol (naming a field/member "stdout").

michaelcfine: You can return an object with two fields if you want to return either a token or an error. E.g., "return { token: null, error: xxx }" or "return { token: xxx, error: null }" and then use these fields here.

michaelcfine (changes requested):
I think the code is looking great... maybe one or two small things. Lots more documentation-related changes proposed, but those should be quick.

docs/Documentation.md
```diff
@@ -58,7 +58,7 @@ Pre- and post- triggers may be defined for any Firebolt method call. See [Trigge
 
 ## novalidate Flag
 
-Use this flag for novalidate mode where Mock Firebolt does not validate uploaded method overrides. This will be used by the certification test suite, not by normal app developer users.
+Use this flag for novalidate mode where Mock Firebolt does not validate options provided by user for inbound or outbound messages. This will be used by the certification test suite, not by normal app developer users.
```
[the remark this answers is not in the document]
vyvanand: Addressed

docs/ProxyMode.md
```diff
@@ -0,0 +1,33 @@
...
+
+To use token while building websocket connection for proxy server and its optional, run below export command before starting Mock Firebolt. 
+
+```export TOKEN=<token>```
+
+If you want to skip validation for specific use case use ```--novalidate``` flag.
+
+```--novalidate method --novalidate params --novalidate response --novalidate event```
+
+*Note*: Proxy mode only work for jsonrpc format.
```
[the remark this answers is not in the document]
vyvanand: we don't wanted to explicitly mention it will work only for thunder messages. So we wanted to mention the thunder message "jsonrpc" format.

docs/ProxyMode.md
```diff
@@ -0,0 +1,33 @@
...
+# Usage
+If you need to use proxy connection for any reason:
+
+```npm run dev -- --proxy 192.168.0.100:9998```
+
+Default port will be 9998 if not passed expliclity.
+
+To use token while building websocket connection for proxy server and its optional, run below export command before starting Mock Firebolt. 
+
+```export TOKEN=<token>```
```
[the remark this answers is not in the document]
vyvanand: changed to MF_TOKEN

server/src/fireboltOpenRpc.mjs
vyvanand: I believe "!" is required here. 
if params is not present in config.validate, it means "--novalidate params" flag is set by user. 
if the flag is set, we shouldn't be validating params for given method and respond with empty array. That's what the method does currently. 

Does it make sense?

server/src/events.mjs
vyvanand: if( config.validate.includes("events") && ! finalResult ) {
        fErr.call(null, method);
      }
Do you mean this?

Error should not be logged even though if "finalResult" is undefined?

basically if user pass in "--novalidate events", config.validate value would be ["method", "response", "params"].
"events" would be removed from default array.

server/src/fireboltOpenRpc.mjs
kschrief: This makes sense to me given the previous discussions. Correct me if I'm wrong that we're maintaining a list of things we need to validate. So "novalidate params" would remove "params" from the validate list.

server/src/events.mjs
kschrief: I agree with Mike on this. The lack of the "event" validation (AKA "! config.validate.includes("events")") should invalidate/remove existing validation functionality. We're merely printing an error here.

Why would an error be printed at this point in the code?

server/src/fireboltOpenRpc.mjs
vyvanand: Correct!

server/src/events.mjs
vyvanand: ok so what would be the validation thats needs to be added against finalResult? @kschrief  @michaelcfine

docs/ProxyMode.md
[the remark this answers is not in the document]
michaelcfine: ah, okay. leave as-is for now. may come back and expand a bit, but at least I understand. thx.

michaelcfine (approved):
👍 Looks good to me!

kschrief (approved):

PR #38, opened 2022-07-12
Firecert 498

apadin002:


server/src/index.mjs
```diff
@@ -59,6 +60,19 @@ server.on('upgrade', function upgrade(request, socket, head) {
   }
 });
 
+
+//Fetch the yaml files from MFOS_Call_Flow repo
+const yamlPath = ".././cli/examples";
+shell.cd(yamlPath)
+try{
+  shell.exec('git clone git+ssh://git@github.comcast.com/lightning-automation/MFOS_Call_Flow')
```
michaelcfine, line 68: We don't want this within Mock Firebolt... it would be a dependency from public github to Comcast's GHE.

michaelcfine (changes requested):
See comment. I think we need to find a different way to do this / clone the repo so that there isn't a dependency from open source GitHub to GHE.

server/src/index.mjs
apadin002: removed the code

apadin002: removed the code

PR #39, opened 2022-07-14
Use ipaddress as userId if userId is missing

naimish15:


server/src/index.mjs
```diff
@@ -42,9 +42,12 @@ server.on('upgrade', function upgrade(request, socket, head) {
   const { pathname } = parse(request.url);
   let userId = pathname.substring(1);
   if ( ! userId ) {
-    logger.info('Using default user');
-    userId = config.app.defaultUserId;
-  } else if ( ! userManagement.isKnownUser(userId) ) {
+    logger.info('Using ip address as userId');
+    let ipAddresses = socket.remoteAddress
+    userId = ipAddresses.slice(ipAddresses.lastIndexOf(":") + 1);
```
kschrief, line 47: This looks like it covers IPv6 addresses only. What about IPv4?

We should be removing all "." in IPv4 addresses and all ":" in IPv6

kschrief (changes requested):

kschrief (approved):

michaelcfine:
I don't think the logic added is quite right, but even if it is, it isn't sufficient... user ids have to 'formally' exist before they can be used (see isKnownUser), so at the very least, you'll need to also add the user with the given userId (IP#). See https://github.com/rdkcentral/mock-firebolt/blob/main/server/src/index.mjs#L63. Note that you have to add the user and then also set up default state for that user (hence the two separate calls).

naimish15:
I was facing some issues in this PR so I will create new PR

PR #41, opened 2022-07-19
Use normalized ip addresses as userId if userId is missing

naimish15:


naimish15:
Will create another PR to add user and state, once we finalize ip values

PR #44, opened 2022-07-26
Feature/GitHub actions

warbhe-amit:
adding github actions for unit test cases


michaelcfine (approved):
I didn't go through every line of every file, but looks okay. One comment: The test for the dot config file should not check against the standard/default/sample config file... the whole point is that users can change their config file to contain whatever apis they want (and change the other settings, too). So it probably only makes sense to test that the file contents are json or something like that.

michaelcfine (dismissed):
I'm not familiar with github actions, but if this all works, looks good to me. We do want to make the unit test check mandatory, though.

kschrief (approved):

warbhe-amit:
@michaelcfine Please review the dotConfig test file
We'll merge this branch once PR 49 gets merged into main.

server/test/suite/dotConfig.test.mjs
```diff
@@ -23,31 +23,6 @@
...
-        cliShortFlag: "d",
-        enabled: false,
-        fileName: "firebolt-discovery-open-rpc.json",
-        name: "discovery",
-      },
-    ],
-    validate: [ "method", "params", "response", "events" ],
-  };
-  expect(dotConfig).toEqual(expectedResult);
+  const result = JSON.parse(JSON.stringify(dotConfig));
```
michaelcfine, line 26: It seems like you are testing that the dotConfig object within the dotConfig module is a valid object. Not sure that's valuable. The code that loads the dotConfig file already has a try-catch when it tries to parse the .mf.config.json file.

So... what's the ultimate point of this test? Do we really need it?

warbhe-amit: Taken care

.github/workflows/npmPublish.yml
```diff
@@ -0,0 +1,33 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
```
kschrief, line 8: This indicates that the job will run on every release. We want the job to be strictly manual for now.

kschrief (changes requested):

.github/workflows/npmPublish.yml
warbhe-amit: added manual configuration on npm publish

kschrief (approved):

kschrief (dismissed):

michaelcfine (dismissed):
👍

kschrief (dismissed):

warbhe-amit:
code coverage is 78.4%

kschrief (approved):

dhillomk (approved):

PR #49, opened 2022-08-09
Feature/refactoring broadcast

warbhe-amit:
code refactoring wrt sendEvent and sendBroadcastEvent


server/src/util.mjs
```diff
@@ -45,6 +47,52 @@ function createTmpFile(prefix, postfix) {
...
+  logger.info(
+    `Could not send ${onMethod} event because no listener is active`
+  );
+};
+
+function fFatalErr() {
+  logger.info(`Internal error`);
+};
+
+function sendEvent(ws, userId, onMethod, result, msg) {
```
michaelcfine, line 66: I see what you're doing here (and with sendBroadcastEvent, but I don't think there should be a dependency from util.mjs to any other file; I'd like to keep it for independent "generic utilities". But you can put these functions inside events.mjs if you want, or just go back to where you're passing in the function pointers, but just pass util.fSuccess, etc.

michaelcfine (changes requested):
Looks good, but I'd like to keep util.mjs free from any dependencies into 'real' code. See comment on util.mjs; either move the utility functions into events.mjs and continue to use or go back to using the full signatures and passing the function pointers, just pass util.fSuccess, etc. (instead of the local anonymous definitions, which you've replaced nicely).

server/src/util.mjs
warbhe-amit: changes done, please check and confirm

michaelcfine (dismissed):
👍 Looks good to me. Thanks for keeping util.mjs nice and simple / dependency-free!

warbhe-amit:
@michaelcfine, rebased with main. Kindly approve it again in order to merge it.

michaelcfine (dismissed):
👍 FWIW, I really don't like the partial imports and then not prefixing the calls (e.g., events.sendEvent vs. sendEvent), but let's leave it as-is.

warbhe-amit:
rebasing with main requires approval

server/src/events.mjs
```diff
@@ -24,9 +24,25 @@ import * as stateManagement from './stateManagement.mjs';
...
 import { config } from './config.mjs';
 
+function logSuccess(onMethod, result, msg) {
+  logger.info(
+    `${msg}: Sent event ${onMethod} with result ${JSON.stringify(result)}`
+  );
+};
+
+function logErr(onMethod) {
+  logger.info(
```
vyvanand, line 37: can this be changed to logger.error?

server/src/events.mjs
```diff
@@ -24,9 +24,25 @@ import * as stateManagement from './stateManagement.mjs';
...
+    `Could not send ${onMethod} event because no listener is active`
+  );
+};
+
+function logFatalErr() {
+  logger.info(`Internal error`);
+};
+
 // Maps full userIds to maps which map event listner request method
 // name (e.g., lifecycle.onInactive) to message id (e.g., 17)
```
vyvanand, line 47: can this be changed to logger.error?

server/src/events.mjs
warbhe-amit: Earlier also this was logger.info only, so if we want to change then should be checked with Mike.

server/src/events.mjs
warbhe-amit: Earlier also this was logger.info only, so if we want to change then should be checked with Mike.

server/src/events.mjs
```diff
@@ -160,28 +176,28 @@ function coreSendEvent(isBroadcast, ws, userId, method, result, msg, fSuccess, f
...
-                  logger.info(`${msg}: Sent event ${method} with result ${JSON.stringify(result)}`)
-                }
-                function fErr() {
-                  logger.info(`Could not send ${method} event because no listener is active`)
-                }
-                function fFatalErr() {
-                  logger.info(`Internal error`)
-                }
-                sendEvent(ws, userId, method, result, msg, fSuccess, fErr, fFatalErr);
+                sendEvent(
```
kschrief, line 179: What's the reason for splitting this up into multiple lines?

server/src/stateManagement.mjs
```diff
@@ -130,6 +130,23 @@ function handleSequenceOfResponseValues(userId, methodName, params, resp, userSt
   return resp;
 }
 
+// Log Error for invalid methodName
+function logError(methodName, resultErrors, resp) {
```
kschrief, line 134: Can we change this name to be more specific to what it's logging?

server/src/triggers.mjs
```diff
@@ -27,6 +27,31 @@ import * as commandLine from './commandLine.mjs';
 const methodTriggers = {};
 const eventTriggers = {};
 
+// Log error
+function logError(errorType, pathDetails, ex) {
```
kschrief, line 31: Can we change this name to be more specific to what it's logging?

kschrief (changes requested):

server/src/events.mjs
warbhe-amit: Taken care

server/src/triggers.mjs
warbhe-amit: Taken care

server/src/stateManagement.mjs
warbhe-amit: Taken care

kschrief (approved):

dhillomk (approved):

PR #52, opened 2022-08-10
Add Discovery Listener; rename Device Listeners

JRicha024:
JIRA:   https://ccp.sys.comcast.net/browse/OTT-1590


cli/examples/device-screenResolution-sequence1.json
```diff
@@ -1,6 +1,6 @@
 {
   "methods": {
-    "device.screenResolution": {
+    "device.onScreenResolutionChanged": {
```
michaelcfine, line 3: These files are meant to say, "Mock Firebolt", please first return this value, then this value, then this value... when someone calls this method. In this case, the goal was for the device.screenResolution call to return these values in order, not to fire events in a sequence. The syntax for a sequence of events is different.

cli/examples/device-screenResolution-sequence2.json
```diff
@@ -1,6 +1,6 @@
 {
   "methods": {
-    "device.screenResolution": {
+    "device.onScreenResolutionChanged": {
```
michaelcfine, line 3: Same comment as above.

cli/examples/device-screenResolution-sequence3.json
```diff
@@ -1,6 +1,6 @@
 {
   "methods": {
-    "device.screenResolution": {
+    "device.onScreenResolutionChanged": {
```
michaelcfine, line 3: Same comment as above.

michaelcfine (changes requested):
See my comments. Please change the screenResolution examples back. If you need/want something that sends a sequence of events, see https://github.com/rdkcentral/mock-firebolt/blob/main/cli/examples/device-onDeviceNameChanged.sequence.json.

cli/examples/device-screenResolution-1.json
```diff
@@ -1,6 +1,6 @@
 {
   "methods": {
-    "device.screenResolution": {
+    "device.onScreenResolutionChanged": {
```
michaelcfine, line 3: This should be just "device.screenResolution"

server/package.json
```diff
@@ -50,6 +50,7 @@
     "express": "^4.17.1",
     "express-handlebars": "^6.0.2",
     "ip": "^1.1.8",
+    "js-yaml": "^4.1.0",
```
michaelcfine, line 53: This doesn't appear to be needed.

michaelcfine (changes requested):
Still found one that needs to go back to the setter method, not the event method. Plus one seemingly spurious entry in the package.json.

cli/examples/device-onNameChanged.sequences.json
```diff
@@ -2,21 +2,21 @@
   {
     "at": 5000,
     "event": {
-      "method": "device.onDeviceNameChanged",
+      "method": "device.onNameChanged",
```
michaelcfine, line 5: One last one... this file should really be called "device-onNameChanged.sequence.json" (no 's' / not plural).

michaelcfine (changes requested):
Just one last minor thing....

server/src/.mf.config.SAMPLE.json
```diff
@@ -1,31 +0,0 @@
-{
```
michaelcfine, line 1: What happened to this file's content?!?!?!

michaelcfine (changes requested):
The .mf.config.SAMPLE.json file seems to have been nuked.

michaelcfine (approved):
👍

michaelcfine (approved):
👍

PR #53, opened 2022-08-16
record request + response and sort by timestamp

vyvanand:


michaelcfine (dismissed):
👍 but curious... won't the records be emitted in time order even without sorting?

vyvanand:
> 👍 but curious... won't the records be emitted in time order even without sorting?

It doesn't. so after 2 request comes in and then response for 1st request been sent back.

michaelcfine (approved):
👍

kschrief (approved):

PR #73, opened 2022-09-23
MF Scope - Implementation

sourabhk2399:
MF Scope - Implementation

### DIFF
--- a/cli/examples/closed-captions-settings.yaml
+++ b/cli/examples/closed-captions-settings.yaml
@@ -4,19 +4,28 @@ methods:
     response: |
       function f(ctx, params) {
         const ccs = ctx.get('closedCaptionsSettings');
-        if ( params && params.hasOwnProperty('value') ) {
-          // Setter called:
-          // Cause an accessibility.onClosedCaptionsSettingsChanged event
-          // (done via post-method trigger on closedcaptions.enabled method)
-          // Save for when accessibility.closedCaptionsSettings is called
-          ccs.enabled = params.value;
-          ctx.set('closedCaptionsSettings', ccs);
-          return true;
-        } else {
-          // Getter called:
-          // Return what we last saved when this func was called as a setter
-          return ccs.enabled;
-        }
+        console.log("### value of ctx inside closedcaptions.enabled of yaml",ctx)
+        console.log("*** value of ccs.enabled in yaml----",ccs.enabled)
+        return ccs.enabled;
+      }
+  closedcaptions.setEnabled:
+    response: |
+      function f(ctx, params) {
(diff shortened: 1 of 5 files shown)


server/src/stateManagement.mjs
```diff
@@ -84,10 +84,23 @@ function addDefaultUser(userId) {
   state[''+userId].isDefaultUserState = true;
 }
 
+// return state based on hierarchy (From lowest priority to highest) global->group->user
 function getState(userId) {
```
kschrief, line 88: So we should be returning a combination of the global/group/user state, not just one of them.

Ex: getState('123~abc') should:
* Load the global state
* Load the state for group "~123" if it exists. All fields in this state should be combined with the global state. Any conflicts should have the group state chosen
* Load the state for userId 'abc~123'. All fields in this state should be combined with the newly created group+global state. Any conflicts should have the userId state chosen

This new state is the one that should be returned

server/src/stateManagement.mjs
```diff
@@ -447,19 +462,52 @@ function mergeCustomizer(objValue, srcValue) {
   }
 }
 
-function updateState(userId, newState) {
-  const userState = getState(userId);
+function updateState(userId, newState, scope = "") {
```
kschrief, line 465: We do not need to add a scope variable here. The userId used will provide us our scope. This is why I'd mentioned that the update functionalities do not need to be updated.

Someone wishing to update the ~abc or "global" scope would just set the userId to "~abc" or "global" when calling the API/CLI. That would then end up here as the userId

kschrief (changes requested):

server/src/stateManagement.mjs
kschrief: After conversation there was a mis-understanding on my part. Resolving conversation.

kschrief (approved):

michaelcfine (approved):
LGTM

kschrief (approved):

cli/examples/closed-captions-settings.yaml
```diff
@@ -4,19 +4,28 @@ methods:
...
-          // Save for when accessibility.closedCaptionsSettings is called
-          ccs.enabled = params.value;
-          ctx.set('closedCaptionsSettings', ccs);
-          return true;
-        } else {
-          // Getter called:
-          // Return what we last saved when this func was called as a setter
-          return ccs.enabled;
-        }
+        console.log("### value of ctx inside closedcaptions.enabled of yaml",ctx)
```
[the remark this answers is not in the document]
sourabhk2399: I'll remove this next commit.

PR #75, opened 2022-10-17
Ott 1650

sourabhk2399:
Fixed a variable issue.

### DIFF
--- a/server/src/stateManagement.mjs
+++ b/server/src/stateManagement.mjs
@@ -95,7 +95,7 @@ function getState(userId) {
   if( userId.includes("~")){
     let group = "~"+userId.split("~")[1];
     if (group in stateCopy){
-      groupState = stateCopy[''+group];
+      let groupState = stateCopy[''+group];
       resetSequenceStateValues(finalState, groupState);
       mergeWith(finalState, groupState, mergeCustomizer);
     }
--- a/server/src/stateManagement.mjs
+++ b/server/src/stateManagement.mjs
@@ -89,24 +89,32 @@ function addDefaultUser(userId) {
 
 // return state based on hierarchy (From lowest priority to highest) global->group->user
 function getState(userId) {
-  const stateCopy = JSON.parse( JSON.stringify(state) )
-  let finalState = stateCopy['global'];
-  userId = '' + userId;
-  if( userId.includes("~")){
-    let group = "~"+userId.split("~")[1];
-    if (group in stateCopy){
-      let groupState = stateCopy[''+group];
-      resetSequenceStateValues(finalState, groupState);
-      mergeWith(finalState, groupState, mergeCustomizer);
+
+  if ( userId in state ) {
+    const stateCopy = JSON.parse( JSON.stringify(state) )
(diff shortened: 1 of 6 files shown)


server/test/suite/stateManagement.test.mjs
```diff
@@ -58,14 +54,14 @@ test(`stateManagement.getAppropriateDelay works properly`, async () => {
     4567,
     "accessibility.closedCaptions"
   );
-  expect(undefinedResultOne).toBeUndefined();
+  expect(undefinedResultOne).toBe(0);
```
kschrief, line 57: Why has this result changed? What's the goal of the test?

server/test/suite/stateManagement.test.mjs
```diff
@@ -58,14 +54,14 @@ test(`stateManagement.getAppropriateDelay works properly`, async () => {
...
 
   stateManagement.testExports.state[6789] = {};
   const undefinedResultTwo = await stateManagement.getAppropriateDelay(
     6789,
     "accessibility.closedCaptions"
   );
-  expect(undefinedResultTwo).toBeUndefined();
+  expect(undefinedResultTwo).toBe(0);
```
kschrief, line 64: Why has this result changed? What's the goal of the test?

server/src/stateManagement.mjs
```diff
@@ -84,10 +87,32 @@ function addDefaultUser(userId) {
...
+        mergeWith(finalState, groupState, mergeCustomizer);
+      }
+    }
+    if (userId in stateCopy){
+      const userState = stateCopy[''+userId];
+      resetSequenceStateValues(finalState, userState);
+      mergeWith(finalState, userState, mergeCustomizer);
+    }
+
+    resetSequenceStateValues(state[''+userId], finalState);
```
kschrief, line 111: How is this functionality different than lines 106-108?

server/test/suite/stateManagement.test.mjs
sourabhk2399: we are checking stateManagement.getAppropriateDelay functionality is properly working or not and currently it is returning 0 as a result so we updated the result with 0.

server/test/suite/stateManagement.test.mjs
sourabhk2399: we are checking stateManagement.getAppropriateDelay functionality is properly working or not and currently it is returning 0 as a result so we updated the result with 0.

server/src/stateManagement.mjs
sourabhk2399: Here I created a copy of state object so that after merging states (global, group and userId), the original global and group does'nt get affected. So first I am making changes in stateCopy ( merging states and handling conflicts).  Then the final state is mereged in the actual sate object and returning it.

kschrief (approved):

michaelcfine (approved):
Looks fantastic. Well-done isolating the functionality into a small number of places (get, set scratch and user state get). Amazing how so much power can be added with relatively few code changes.

PR #87, opened 2022-12-07
OTT-1758-update device module in group scope

Nandana-NNR:


scthunderbolt:

PR #99, opened 2022-12-23
Ott 1683 2

kschrief:


dhillomk (approved):

dhillomk (dismissed):

server/src/stateManagement.mjs
```diff
@@ -80,23 +80,115 @@ addDefaultUser(config.app.defaultUserId);
...
 }
 
+//returns userId with given user/appId if present
+function getUserId(userId){
+  var users = Object.keys(state);
+
+  // //Checking if user or appId is present in state object
+  if (!(userId in state)){
+    for(var key in users){
+      if (users[key].includes("~")){
```
michaelcfine, line 159: I didn't look super closely but this code looks a lot like the code above (for parsing the full user id. I think we should have utility function(s) for this to keep the code DRY.

server/src/userManagement.mjs
```diff
@@ -117,6 +119,59 @@ function heartbeat(ws) {
 }
 
 function addUser(userId) {
+  userId = "" + userId;
+  var users = getUsers();
+  let appId,user,group;
+
+  //getting user, group and appId from userId
+  if (userId.includes("~")){
```
michaelcfine, line 127: Same comment as above in terms of seemingly duplicative userid parsing code.

michaelcfine (commented):
Looks good other than (a) I think we should keep the POST to create a user without an ID and just add the PUT next to it; and (b) there looks to be duplicative userid parsing code that looks like it could be refactored.

server/src/stateManagement.mjs
kschrief: Added a utility function that returns an object containing the split userId as user, group, and/or appId

server/src/userManagement.mjs
kschrief: Added a utility function that returns an object containing the split userId as user, group, and/or appId

cli/README.md
```diff
@@ -61,6 +61,17 @@ To process json file containing sequence of events :
 run `node cli.mjs --sequence "path of sequence.json"`
 example : node cli.mjs --sequence ../examples/device-onDeviceNameChanged.sequence.json
 
+## Status of ws connection
+
+To get the status of web-socket connection of user :
+run `node cli.mjs --getStatus`
```
michaelcfine, line 67: I think this is probably fine, but now that I think about it, it might be nice if the name of this cli param included what the status is for/about. E.g., getUserStatus? getSocketStatus?

cli/README.md
```diff
@@ -61,6 +61,17 @@ To process json file containing sequence of events :
...
+
+## Adding user
+
+To add user :
+run `node cli.mjs --addUser <userId>`
+example: node cli.mjs --addUser "123~A#netflix"
```
michaelcfine, line 73: Did we land where we support both POST and PUT? If so, do we have two separate CLI params/commands or just one with an optional userId value? If the former, we should document both. If the latter, we should probably put square brackets around userId and show two examples, one with a user id and one without (with a note that a UUID will be generated).

cli/src/cli.mjs
```diff
@@ -138,6 +140,19 @@ if ( parsed.help ) {
 
   usage();
 
+} else if ( parsed.addUser ) {
+
+  const user = parsed.addUser;
+
+  msg(`adding user ${user}`);
+  axios.put(url(host, port, `/api/v1/user/${ encodeURIComponent(user)}`), undefined)
```
michaelcfine, line 148: Not sure if it's on a different branch, but assuming we have both POST and PUT support, I would assume this would test whether a user id value was given and do one or the other (assuming we only have one CLI command/param).

docs/MFScope.md
```diff
@@ -4,23 +4,32 @@ This feature allows user to share state belonging to same group.
...
 
-- A user without a group will contain simply the user: "foo"
-- A user with a group will contain both a user and group separated by "~". Ex: "foo~bar" represents user "foo" who is a part of group "bar"
-- A group without a user will contain "~" before the group ID. Ex: "~bar".
-    - This example represents group "bar" only.
-    - A userId containing just "bar" will be interpreted as the user "bar" that is not a part of a group.
+| Full "userId"   | User | Group | AppId   |
+| --------------- | ---- | ----- | ------- |
+| 123             | 123  | N/A   | N/A     |
+| 123~A           | 123  | ~A    | N/A
```
michaelcfine, line 17: Does this need a final pipe character?

docs/MFScope.md
```diff
@@ -4,23 +4,32 @@ This feature allows user to share state belonging to same group.
...
+| 123~A#netflix   | 123  | ~A    | netflix |
+| 123#netflix     | 123  | N/A   | netflix |
 
 There will also be a "global" user. This is a reserved userId representing the global state across all users and groups.
 
+All users and appId are unique:
```
michaelcfine, line 23: nit: appIds, plural.

docs/MFScope.md
```diff
@@ -4,23 +4,32 @@ This feature allows user to share state belonging to same group.
...
+- Two users with same user part can not exist (if "123~A" exist, "123~B" can not exist).
```
michaelcfine, line 24: nit: cannot is one word.

docs/MFScope.md
```diff
@@ -87,6 +99,26 @@ curl --location --request PUT 'http://localhost:3333/api/v1/state' \
 	    }
 	}'
 ```
+
+- **To set state for "123~A#netflix" user via HTTP, use below curl command :**
+
+```
+curl --location --request PUT 'http://localhost:3333/api/v1/state' \
+	--header 'content-type: application/json'  \
+--header 'x-mockfirebolt-userid: 123~A#netflix'  \
```
michaelcfine, line 108: Nit: Should this be indented? Layout/formatting looks strange.

server/README.md
```diff
@@ -494,3 +511,28 @@ curl --location --request GET 'http://localhost:3333/api/v1/user'
     ]
 }
 ```
+
+## Get ws status (returns the status of ws connection mapped to the userID)
+
+### Example cURL Command:
+
+```
+curl --location --request GET 'http://localhost:3333/api/v1/status'
```
michaelcfine, line 520: Does it make sense to show an example where you provide the userId value in the header? And/or make it clear the default user will be used (like for all other commands)?

server/README.md
```diff
@@ -494,3 +511,28 @@ curl --location --request GET 'http://localhost:3333/api/v1/user'
...
+```
+
+### 200 Response (If connection is established):
+
+```
+{
+   status": "WS connection found",
+   "readyState": "CONNECTED"}
```
michaelcfine, line 528: nit: ending brace is probably not right, I'm guessing.

server/README.md
```diff
@@ -462,7 +462,7 @@ curl --location --request POST 'http://localhost:3333/api/v1/sequence' \
 ### Example cURL Command:
 
 ```
-curl --location --request POST 'http://localhost:3333/api/v1/user'
+curl --location --request PUT 'http://localhost:3333/api/v1/user'
```
michaelcfine, line 465: Can we show both put and post?

server/src/routes/api/user.mjs
```diff
@@ -27,21 +27,49 @@ import * as stateManagement from '../../stateManagement.mjs';
...
-// POST /api/v1/user
+// PUT /api/v1/user/:userId?
+// POST /api/v1/user/:userId?
 // Expected body: N/A
+// URL Param "userId" - Optional
 function addUser(req, res) {
-  // Generate a unique userId for this user (no vanity userIds, at least for now)
-  const userId = uuidv4();
+  let userId;
+  if (req.params.userId){
```
michaelcfine, line 36: It feels like this function could be refactored to avoid dup code. Just something up front that either uses the given user id if given or generates one. Then the core logic that does the stateManagement.addUser, userManagement.addUser, and returns success or error.

michaelcfine (changes requested):

server/src/routes/api/user.mjs
```diff
@@ -27,21 +27,47 @@ import * as stateManagement from '../../stateManagement.mjs';
 
 // --- Route Handlers ---
 
-// POST /api/v1/user
+// PUT /api/v1/user
```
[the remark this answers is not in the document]
kschrief: Support has been added for PUT and POST now

cli/README.md
kschrief: Changed to "--getSocketStatus" in the CLI and all documentation. I left the API alone as "/api/v1/status"

docs/MFScope.md
kschrief: Weirdly enough no. But it has one now anyway.

PR #100, opened 2023-01-03
support concurrent proxy connection

vyvanand:
initializeAndSendRequest method will open new websocket for each request.
Once response received from proxy, it will start closing all stale connections.
(ex: if 10 connections open for 10 request, closeStaleConnection method will close the WS connection after interval of 10s one after other. )
This way, it will execute post trigger and stale connections won't be open all the time.


kschrief:
Closed PR. We will not be merging in this branch.

PR #101, opened 2023-01-05
OTT-1785

ksentak:
Added an additional utility function that allows WS to listen to a single event that is fired off once it is added to a group that is actively listening. Also added additional test cases to ensure that correct response is received after triggering a cli event. The reasoning behind creating a separate function that is different than  `fireboltCommand`  is that although the code is similar, I anticipate needing to build out additional functionality within the new `mfEventListener` function to be able to listen to more that a single event before closing the WS connection.


functional/tests/utilities.mjs
```diff
@@ -223,4 +223,39 @@ async function mfState(on, extraConfig = "") {
...
+ * @param {Number} port ws port to connect
+ * @param {String} user to pass the userID
+ * @param {Number} eventCommands how many commands will be sent once eventListener is active
+ * @returns Promise yielding the response on resolve()
+ */
+const mfEventListener = (command, port, user) => {
+  return new Promise((resolve) => {
+    const wsClientURL = `${wsClient}${port || 9998}${user ? `/${user}` : ''}`;
+    const ws = new WebSocket(wsClientURL);
+    ws.on("open", () => {
```
kschrief, line 237: Do we want to put this after "message"? That way we've properly defined how to receive the response _before_ we send the message across?

functional/tests/utilities.mjs
```diff
@@ -223,4 +223,39 @@ async function mfState(on, extraConfig = "") {
...
+      ws.send(command);
+    });
+
+    const eventMap = new Map();
+
+    ws.on("message", (data) => {
+      const dataObj = JSON.parse(data);
```
kschrief, line 244: Let's do some error handling here. Not much we can do other than log the error and drop the message but that's a cleaner fail.

kschrief (changes requested):

PR #102, opened 2023-01-27
Fix/scoping scenarios

kschrief:


server/src/stateManagement.mjs
```diff
@@ -201,9 +202,27 @@ function getState(userId) {
       }
     }
     if (userId in stateCopy){
+      let appId, user
       const userState = stateCopy[''+userId];
-      resetSequenceStateValues(finalState, userState);
-      mergeWith(finalState, userState, mergeCustomizer);
+      user = userId.split("~")[0];
```
kschrief, line 207: There is a function called "parseUser" in userManagement that will automatically split the userId into an object containing user, group, and appId. Please use this to avoid duplicating code

server/src/stateManagement.mjs
```diff
@@ -201,9 +202,27 @@ function getState(userId) {
...
+      if (userId.includes("#")){
+        appId = userId.split("#")[1];
+      }
+      /**if userId is updated via user scope(either by user/appId/full userid),
+       * it would be pushed to userIdArray when updateState() was triggered, here preference is given to userScope
+       */
+      if (userIdArray.includes(user) || userIdArray.includes(appId) || userIdArray.includes(userId) == true) {
```
kschrief, line 214: Hmm. We shouldn't have a user entry for the appId directly. We should be checking users to see if a user contains that appId. Is that not what's happening?

PR #103, opened 2023-01-30
Fix/scoping scenarios

kschrief:


dhillomk (approved):

michaelcfine (approved):
I had a hard time understanding things here, in part because I wasn't sure what "trackUpdateState" was used for. I _think_ it's just for testing, not 'core' logic. I encourage you to add a comment where the variable is declared explaining what it is / what it's for / when/why it gets updated, how it's used.

Also, there's a ton of non-DRY code related to the parsing of full user ids and ensuring user id values are strings (''+userId). If we don't already have a ticket for this, we should. I think a single utility function that takes a userId and returns the three components (optionally) (or a small number of utilities, as needed) would go a long way to simplifying and DRYing the code.

scthunderbolt:

PR #106, opened 2023-02-02
SYSTEST-9245-Log user IDs in an array in MF logs instead of multiple lines

Nandana-NNR:


server/src/index.mjs
```diff
@@ -82,16 +82,16 @@ server.on('upgrade', function upgrade(request, socket, head) {
 });
 
 // Starter user(s)
-logger.info('Adding user 123...'); stateManagement.addUser('123'); userManagement.addUser('123');
-logger.info('Adding user 456...'); stateManagement.addUser('456'); userManagement.addUser('456');
-logger.info('Adding user 789...'); stateManagement.addUser('789'); userManagement.addUser('789');
+stateManagement.addUser('123'); userManagement.addUser('123');
```
kschrief, line 85: Can you remove these three lines? With the appId changes adding 123, 456, 789 is preventing 123~A, 456~A, and 789~A from being added which are commonly used userIds.

server/src/routes/api/user.mjs
```diff
@@ -23,6 +23,8 @@
 import { v4 as uuidv4 } from 'uuid';
 import * as userManagement from '../../userManagement.mjs';
 import * as stateManagement from '../../stateManagement.mjs';
+import { logger } from '../../../src/logger.mjs';
```
kschrief, line 26: You don't have to go all the way up to root. You can just ../../logger.mjs

server/src/stateManagement.mjs
```diff
@@ -738,6 +742,7 @@ export const testExports={
 }
 export {
   state,
+  defaultUserArray,
```
kschrief, line 745: We can just call userManagement.getUsers(). We don't need to keep track of "default" users

server/src/routes/api/user.mjs
```diff
@@ -47,6 +49,7 @@ function addUser(req, res) {
         status: 'SUCCESS',
         userId: userId
       });
+      logger.info(`Adding user ${userId}`);
```
kschrief, line 52: Can we say "Added" instead of "Adding"?

kschrief (changes requested):

dhillomk (approved):

PR #108, opened 2023-02-06
Update manage SDK version to *

kschrief:


michaelcfine (approved):
I think this is fine for now, at least until we decide to formally move SDK dependencies out of package.json and into the config file (and making them runtime dependencies rather than package.json/build-time dependencies).

dhillomk (approved):

PR #111, opened 2023-02-14
Update userManagement.mjs

kdivya153:
Added logic 1:many

### DIFF
--- a/server/src/userManagement.mjs
+++ b/server/src/userManagement.mjs
@@ -132,11 +132,12 @@ function associateUserWithWs(userId, ws) {
   const wsArray=[];
   if(user2ws.has(''+userId)){
     wsArray = user2ws.get(''+userId)
-    wsArray.push(ws);
+    wsArray.push([ws]);
   } else {
-    wsArray.push(ws)
+    wsArray.push([ws])
     user2ws.set(''+userId, wsArray);
   }
+  logger.info("User have associated with ws")
 }
 
 
--- a/server/src/userManagement.mjs
+++ b/server/src/userManagement.mjs
@@ -132,7 +132,7 @@ function associateUserWithWss(userId, wss) {
 function associateUserWithWs(userId, ws) {
   let wsArray = [];
   if(user2ws.has(''+userId)){
-    wsArray = user2ws.get(''+userId)
+    wsArray = [user2ws.get(''+userId)]
     wsArray.push(ws);
   } else {
     wsArray.push(ws)
--- a/server/src/userManagement.mjs
+++ b/server/src/userManagement.mjs
@@ -86,9 +86,8 @@ function getWssForUser(userId) {
 
 //get the ws for user,if user has ws already,it will return the latest ws connection
 function getWsForUser(userId) {
-  let wsArray = [];
   if ( user2ws.has(''+userId) ) {
-    wsArray=user2ws.get(''+userId);
+    let wsArray=user2ws.get(''+userId);
     let latestWsConnection=wsArray[(wsArray.length-1)]
     return latestWsConnection;
   }
@@ -130,7 +129,7 @@ function associateUserWithWss(userId, wss) {
 
 //Associate each user with many ws connection,taking ws as an array of objects
 function associateUserWithWs(userId, ws) {
-  let wsArray = [];
+  const wsArray=[];
   if(user2ws.has(''+userId)){
     wsArray = user2ws.get(''+userId)
     wsArray.push(ws);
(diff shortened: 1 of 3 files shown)


jbigel (dismissed):

dhillomk (dismissed):

jbigel (dismissed):

server/src/userManagement.mjs
```diff
@@ -84,9 +84,13 @@ function getWssForUser(userId) {
   return undefined;
 }
 
+//get the ws for user,if user has ws already,it will return the latest ws connection
 function getWsForUser(userId) {
+  let wsArray = [];
```
michaelcfine, line 89: Nit: You don't need this here since you only use wsArray in the if block. Or if you do, you don't have to assign it.

server/src/userManagement.mjs
```diff
@@ -84,9 +84,13 @@ function getWssForUser(userId) {
...
   if ( user2ws.has(''+userId) ) {
-    return user2ws.get(''+userId);
+    wsArray=user2ws.get(''+userId);
+    let latestWsConnection=wsArray[(wsArray.length-1)]
```
michaelcfine, line 92: Can/should use 'const' here.

server/src/userManagement.mjs
```diff
@@ -124,10 +128,19 @@ function associateUserWithWss(userId, wss) {
   user2wss.set(''+userId, wss);
 }
 
+//Associate each user with many ws connection,taking ws as an array of objects
 function associateUserWithWs(userId, ws) {
-  user2ws.set(''+userId, ws);
+  let wsArray = [];
```
michaelcfine, line 133: Same comment as above; Can either define this within the if block or just not assign a value here.

server/src/userManagement.mjs
```diff
@@ -124,10 +128,19 @@ function associateUserWithWss(userId, wss) {
...
+  if(user2ws.has(''+userId)){
+    wsArray = user2ws.get(''+userId)
+    wsArray.push(ws);
+  } else {
+    wsArray.push(ws)
```
michaelcfine, line 138: These two lines are fine, but could be replaced by:
wsArray.push([ ws ]);

server/src/userManagement.mjs
michaelcfine: More importantly, is this the right "logic"? Do we always want to return the last ws object in the array? Is there going to be code that removes items from this array as connections disconnect? It feels like there's something missing here.

michaelcfine (commented):
See comments in-line. Main thing: the logic for get is always to return the last item in the array. Is that what we really want? Is there some other logic missing somewhere else?

server/src/userManagement.mjs
kdivya153: Moved the code inside if block,It is expecting me to assign it explicitly as an array.else i'm getting the below error.
 **TypeError: Cannot read properties of undefined (reading 'push')**

server/src/userManagement.mjs
kdivya153: Declared wsArray as const.

kdivya153: As per the ask,changes should support the backward compatibility, which has to return the last ws object in the array.I'm suspecting user2ws.delete() will delete the user along with the list of ws connections in the array once connection disconnect.

server/src/userManagement.mjs
kdivya153: wsArray is used in both if as well else block.Declared it as const

server/src/userManagement.mjs
kdivya153: wsArray.push([ws]),places array of ws inside an array,So when userManagement.getWsForUser is called,It is retrieving the values as [test] and validation steps are getting failed,because the expected value is "test''

server/src/userManagement.mjs
michaelcfine: Hmm... const should be allowed when the value of a variable doesn't change, but it's okay with let.

server/src/userManagement.mjs
michaelcfine: whoops... actually meant wsArray = [ ws ]; but what you have is fine.

michaelcfine (dismissed):
👍

jbigel (approved):

server/src/userManagement.mjs
kdivya153: done

dhillomk (approved):

jbigel (approved):

dhillomk (approved):

kschrief (approved):

michaelcfine (commented):
👍

PR #113, opened 2023-03-01
Systest 9252

kdivya153:


kschrief (approved):

server/src/stateManagement.mjs
```diff
@@ -179,38 +178,38 @@ function getUserId(userId){
   return userId;
 }
 
-// return state based on hierarchy (From lowest priority to highest) global->group->user
-function getState(userId) {
+// return state based on hierarchy (From lowest priority to highest) global->group->user if mergedState=true
+//return state of the UserId if mergedState=false,to update the state
```
michaelcfine, line 182: This comment seems confusing; I wouldn't expect getXxx to update. Is this just a cut-and-pasteo?

server/src/stateManagement.mjs
```diff
@@ -179,38 +178,38 @@ function getUserId(userId){
...
-      if (group.includes("#")){
-        group = group.split("#")[0];
+  if(!mergedState){
+    return state[''+userId];
+  }else{
+    if ( userId in state ) {
+      const stateCopy = JSON.parse( JSON.stringify(state) )
+      let finalState = stateCopy['global'];
+      userId = '' + userId;
+      if( userId.includes("~")){
```
michaelcfine, line 194: Can we please encapsulate all the "parsing" of userIds into coreUserId, group, appId into one utility function (or set of utility functions) so that code that's doing higher order things don't have to be cluttered by that kind of lower-level parsing logic?

server/src/stateManagement.mjs
```diff
@@ -584,13 +583,12 @@ function updateState(userId, newState, scope = "") {
   }
   //to get the userId from given user/appId
   scope = getUserId(scope);
-
   if ( scope in state ){
-    userState = getState(scope);
+    userState = getState(scope,false);
```
michaelcfine, line 587: It seems strange to use 'scope' as a variable name that holds an actual user id (line 581). I think of 'scope' as either the literal string 'global' or the literal string 'group' or the literal string 'user' or something like that.

michaelcfine (commented):
This code seems cleaner and closer to what I expected/expect, but (a) we need to isolate the user id string parsing into one or a small number of utility functions and get the parsing code out of the main logic; (b) we should try to not need ''+userId as much somehow; (c) the code still feels different than I expected. But if Kurt or others closer to this are okay with the main logic, I'm okay with it once we isolate the userId parsing code.

server/src/stateManagement.mjs
kdivya153: @michaelcfine,I have used the existing parse(userId) in userManagment.mjs,to parse the userId and get appId, group,then assigned those values to the finalstate object.

server/src/stateManagement.mjs
```diff
@@ -28,7 +28,7 @@ import * as fireboltOpenRpc from './fireboltOpenRpc.mjs';
 import * as commonErrors from './commonErrors.mjs';
 import * as util from './util.mjs';
 import { sendBroadcastEvent, sendEvent, logSuccess, logErr, logFatalErr } from './events.mjs';
-import { v4 as uuidv4 } from 'uuid';
+import { parse, v4 as uuidv4 } from 'uuid';
```
kschrief, line 31: Looks like VS Code thought you were gonna use this but I don't see it anywhere. Can we get rid of it?

server/src/stateManagement.mjs
```diff
@@ -179,38 +178,37 @@ function getUserId(userId){
...
+    if ( userId in state ) {
+      const stateCopy = JSON.parse( JSON.stringify(state) )
+      let finalState = stateCopy['global'];
+      //Parsing the UserId to get group and appid
+      let parseUserId=parseUser(userId)
+      if(!(parseUserId.group == undefined && parseUserId.appId== undefined))
+      {
+        if(parseUserId.group || parseUserId.appId in stateCopy)
+        {
+          let groupOrAppState = parseUserId.appId ? stateCopy['#'+parseUserId.appId]:stateCopy['~'+parseUserId.group];
```
kschrief, line 199: Why are we dealing with app state here? Getting the user state for an appId is done in getUserId() on line 186. We should only need to reference the group here.

server/src/userManagement.mjs
```diff
@@ -44,6 +44,14 @@ function getUsers() {
 function isKnownUser(userId) {
   return user2wss.has(''+userId);
 }
+/* 
+* @function:parseUser()
+* @Description: parsing the userId to get user,group and appId
+* @Params:userId
+* @Return: Output object
+* @Example Input: 123~A#youtube
```
kschrief, line 52: The appIds need to be generic. Can this be "appId1" instead?

Also thank you for adding this. :)

kschrief (changes requested):

server/src/userManagement.mjs
kdivya153: Done.

server/src/stateManagement.mjs
kdivya153: As per the previous existing logic, If the group exist in the userId, we are assigning the group to the "group" variable and it it getting overridden when the appId exist in the userId.So I have rewritten the same logic, to "groupOrAppState"variable will have value as appId if it is exist in the userId, else it will have group. 

Here is the previous existing code before change:

```if( userId.includes("~")){
      let group = "~"+userId.split("~")[1];    ---//group variable is assigned to the group in userId
      if (group.includes("#")){
        group = group.split("#")[0];  ----------//group variable is assigned to the appId in userId
      }
      if (group in stateCopy){
        let groupState = stateCopy[''+group];
        logger.info(`Group State in grp of user${userId} ${JSON.stringify(groupState)}`);
        resetSequenceStateValues(finalState, groupState);
        mergeWith(finalState, groupState, mergeCustomizer);
        logger.info(`Final state in grp of user${userId} ${JSON.stringify(finalState)}`);
      }

Do you want me to change like below?
 if(!(parseUserId.group == undefined)
      {
        if(parseUserId.group in stateCopy)
        {
          let groupOrAppState = stateCopy['~'+parseUserId.group];
          resetSequenceStateValues(finalState, groupOrAppState);
          mergeWith(finalState, groupOrAppState, mergeCustomizer);
        }
    }

kdivya153: Done with the changes.

server/src/userManagement.mjs
kschrief: I meant the "#youtube" in the example needs to become "#appId1. We don't need to change the variable name since it's already generic.

kschrief (changes requested):

server/src/userManagement.mjs
kdivya153: Done

kschrief (approved):

jbigel (approved):

PR #114, opened 2023-03-07
SYSTEST-9249-Set warning modes for multiple connections for a single user

Nandana-NNR:


server/test/suite/config.test.mjs
```diff
@@ -24,22 +24,23 @@ import { config } from "../../src/config.mjs";
 
 test(`config works properly`, () => {
   const expectedResult = {
+    validate: ["method", "params", "response", "events"],
```
kschrief, line 27: It looks like there are a large number of unnecessary changes in this file.

server/src/commandLine.mjs
```diff
@@ -89,6 +90,11 @@ const developerToolName = parsed.developerToolName || config.app.developerToolNa
 const proxy = parsed.proxy;
 const conduit = parsed.conduit;
 
+// Overriding multiUserConnections, if a configurartion is passed via CLI on staring MF server
+// Else default configuration 'warn' is used
+const multiUserConnections = parsed.multiUserConnections;
+config.multiUserConnections = multiUserConnections ? multiUserConnections : config.multiUserConnections
```
kschrief, line 96: Let's also validate the config here as well.

If the user provided config is not "allow", "warn" or "deny" we default to what's in the configuration file (warn). Those comparisons should be caps insensitive.

server/src/userManagement.mjs
```diff
@@ -228,11 +228,21 @@ function addUser(userId) {
     ws.on('pong', async hb => {
       heartbeat(ws)
     });
-    associateUserWithWs(''+userId, ws);
-    handleGroupMembership(''+userId)
-    ws.on('message', async message => {
-      messageHandler.handleMessage(message, ''+userId, ws);
-    });
+    if ((config.multiUserConnections == 'allow') || (config.multiUserConnections == 'warn') || (config.multiUserConnections == 'deny' && getWsForUser(userId) == undefined)) {
```
kschrief, line 231: We should clean up this logic a bit. It looks like it will give the warning log even on the first connection?

Maybe we can put the denial logic at the beginning to get that out of the way. Let's also close() the incoming ws connection so the client knows we're not accepting the connection.

After that we can give a warning log if needed but either way let the connection through.

server/src/userManagement.mjs
```diff
@@ -228,11 +228,21 @@ function addUser(userId) {
...
+      if (config.multiUserConnections == 'warn') {
+        logger.importantWarning(`WARNING: We do not support multiple websocket connections for a single user`)
```
kschrief, line 233: This can be a bit more descriptive. Something like

"Mock Firebolt was not written to support multiple connections for a single user. Some advanced use cases may result in unexpected behaviors."

kschrief (changes requested):

server/test/suite/config.test.mjs
Nandana-NNR: Hey Kurt, the only actual change in this file is line no: 28 and line no:43, wherein the test case was updated for including our new configuration - multiUserConnections: "warn"
The rest of the changes are just re-arranging the existing kv pairs a bit (no existing value has been updated/deleted), if you take a closer look
The reason is when I run this test locally, the test passes with just the changes in 28 and 43. But when the same test was run from concourse as part of PR, the response of 'config' had the fields differing in order, hence changed the order in the test

server/src/commandLine.mjs
```diff
@@ -89,6 +90,13 @@ const developerToolName = parsed.developerToolName || config.app.developerToolNa
 const proxy = parsed.proxy;
 const conduit = parsed.conduit;
```
Nandana-NNR, line 92: **Note:** test() is a JS inbuilt RegExp method to test for a match in a string. Adding a note here to avoid any confusions
Documentation for reference- https://www.w3schools.com/jsref/jsref_regexp_test.asp

server/test/suite/config.test.mjs
kschrief: No worries. I knew it was mostly formatting but I didn't know there was a reason behind some of it.

kschrief (dismissed):

kschrief (dismissed):

server/src/commandLine.mjs
```diff
@@ -89,6 +90,13 @@ const developerToolName = parsed.developerToolName || config.app.developerToolNa
...
 
+// Overriding multiUserConnections, if a configurartion, eg: allow/warn/deny (case-insensitive) is passed via CLI on starting MF server
+// Else default configuration 'warn' is used
+const multiUserConnections = parsed.multiUserConnections;
+if(/allow/i.test(multiUserConnections) || /warn/i.test(multiUserConnections) || /deny/i.test(multiUserConnections) == true){
```
michaelcfine, line 96: "== true" is usually a strange formulation in JS given how it treats 'truthiness'. It's okay, but you probably can just remove it if I understand what you're trying to do.

server/src/userManagement.mjs
```diff
@@ -236,11 +236,22 @@ function addUser(userId) {
     ws.on('pong', async hb => {
       heartbeat(ws)
     });
-    associateUserWithWs(''+userId, ws);
-    handleGroupMembership(''+userId)
-    ws.on('message', async message => {
-      messageHandler.handleMessage(message, ''+userId, ws);
-    });
+    // If multiUserConnections configuration is set as deny and there is a ws object associated with userId, deny and log second ws connection and drop the attempt
+    if (/deny/i.test(config.multiUserConnections) == true && getWsForUser(userId) !== undefined) {
```
michaelcfine, line 240: "!== undefined" is fine if you're *sure* it will always be undefined when, uh, undefined and not, say, null. A slightly less strict check would simply be to test the falsiness of getWsForUser(userId)... E.g., just use "!getWsForUser(userId)" instead.

server/src/userManagement.mjs
```diff
@@ -236,11 +236,22 @@ function addUser(userId) {
...
+      logger.info(`Denying second websocket connection of user ${userId}`)
+      ws.close();
+    }
+    else {
+      // Else, If multiUserConnections configuration is set as warn, and there is a ws object associated with userId, warn and allow connection
+      if (/warn/i.test(config.multiUserConnections) == true && getWsForUser(userId) !== undefined) {
+        logger.importantWarning(`WARNING: Mock Firebolt was not written to support multiple connections for a single user. Some advanced use cases may result in unexpected behaviors`)
```
michaelcfine, line 247: Nit: Add a period at the end of the last sentence.

michaelcfine (approved):
Looks good. A few very minor suggestions. Approving if you can't make these minor changes or just need to move onto whatever's higher priority.

server/src/commandLine.mjs
Nandana-NNR: Addressed

server/src/userManagement.mjs
Nandana-NNR: Addressed

server/src/userManagement.mjs
Nandana-NNR: getWsForUser method was implemented in such a way it always returns 'undefined' if there is no ws object associated with a userid, hence the check
```
    let wsArray=user2ws.get(''+userId);
    let latestWsConnection=wsArray[(wsArray.length-1)]
    return latestWsConnection;
  }
  return undefined;
}
```

michaelcfine (approved):
👍

kschrief (approved):

PR #115, opened 2023-03-07
SYSTEST-9250 1-1 WS Mapping for Proxy Mode

ksentak:
Create 1-1 WS mapping instead of a singleton WS connection when MF is in proxy mode.

### DIFF
--- a/server/src/messageHandler.mjs
+++ b/server/src/messageHandler.mjs
@@ -280,7 +280,7 @@ async function handleMessage(message, userId, ws) {
 
   logger.debug(`Sending response for method ${oMsg.method}`);
   let finalResponse = (newResponse ? newResponse : response);
-  if (!process.env.proxy) {
+  if (!process.env.proxy || stateManagement.hasOverride(userId, oMsg.method)) {
     const oResponseMessage = {
       jsonrpc: '2.0',
       id: oMsg.id,
--- a/server/src/messageHandler.mjs
+++ b/server/src/messageHandler.mjs
@@ -161,7 +161,7 @@ async function handleMessage(message, userId, ws) {
     //bypass JSON-RPC calls and hit proxy server endpoint
     //init websocket connection for proxy request to be sent and use receiver client to send events back to caller.
     try {
-      response = await proxyManagement.initializeAndSendRequest(ws, JSON.stringify(oMsg))
+      response = await proxyManagement.sendRequest(ws, JSON.stringify(oMsg))
     } catch (err) {
       logger.error(`ERROR: Unable to establish proxy connection due to ${err}`)
       process.exit(1)
--- a/server/src/messageHandler.mjs
+++ b/server/src/messageHandler.mjs
@@ -161,12 +161,7 @@ async function handleMessage(message, userId, ws) {
     //bypass JSON-RPC calls and hit proxy server endpoint
     //init websocket connection for proxy request to be sent and use receiver client to send events back to caller.
     try {
-      if(await proxyManagement.initialize(proxyManagement.actOnResponseObject, ws)) {
-        proxyManagement.sendRequest(JSON.stringify(oMsg))
-        response = await proxyManagement.getResponseMessageFromProxy(oMsg.id)
-      } else {
-        console.log("Websocket connection not initialized")
-      }
+      response = await proxyManagement.initializeAndSendRequest(ws, JSON.stringify(oMsg))
     } catch (err) {
       logger.error(`ERROR: Unable to establish proxy connection due to ${err}`)
       process.exit(1)
--- a/server/src/proxyManagement.mjs
+++ b/server/src/proxyManagement.mjs
@@ -24,7 +24,7 @@ import WebSocket from 'ws';
 const wsMap = new Map();
 const wsMsgMap = new Map();
 
-async function initializeAndSendRequest(returnWs, command) {
+async function sendRequest(returnWs, command) {
   let outgoingWs = wsMap.get(returnWs);
   /* Checks to see if ws connection is in map.
    * If connection exists and is active, it will be used.
@@ -154,4 +154,4 @@ function getMFToken(request) {
 }
 
 // --- Exports ---
-export { getMFToken, initializeAndSendRequest };
+export { getMFToken, sendRequest };
--- a/server/src/proxyManagement.mjs
+++ b/server/src/proxyManagement.mjs
@@ -22,19 +22,17 @@ import { parse } from 'url';
 import WebSocket from 'ws';
 
 const wsMap = new Map()
-let closeConnectionTrigger = false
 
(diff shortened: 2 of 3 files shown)


server/src/proxyManagement.mjs
```diff
@@ -21,97 +21,114 @@
...
+  outgoingWs.send(command);
+
+  return new Promise(async (res, rej) => {
+    try {
+      const response = await getResponseMessageFromProxy(returnWs);
+      // Clear for next message
+      wsMsgMap.set(returnWs, null);
+      res(response);
+    } catch (err) {
+      rej('Timeout occurred');
```
kschrief, line 60: Can we make this message a little more specific? Something like "Timeout waiting for WS repsonse"

server/src/proxyManagement.mjs
```diff
@@ -21,97 +21,114 @@
...
+
+function setupOutgoingWs(returnWs) {
+  const url = buildWSUrl();
+  const ws = new WebSocket(url);
   try {
     return new Promise((res, rej) => {
       ws.on('open', function open() {
-        console.log("Connection to websocket proxy server established")
-        res(true)
+        console.log('connection established');
```
kschrief, line 71: Can we keep the old log message? "Connection to websocket proxy server established"

server/src/proxyManagement.mjs
```diff
@@ -21,97 +21,114 @@
...
       }
-      if(responseObject[id]) {
-        counter = timeout + interval
-        //clear interval if response received for given id.
-        clearInterval(timer)
-        resolve(responseObject[id])
+
+      const returnMsg = wsMsgMap.get(returnWs);
+      if (returnMsg) {
+        counter = timeout + interval;
```
kschrief, line 105: Why are we updating the interval here even though we're done?

server/src/proxyManagement.mjs
```diff
@@ -21,97 +21,114 @@
...
-  let proxyUrl = process.env.proxyServerIP
-  if( ! proxyUrl ) {
-    throw Error('ERROR: Proxy Url not found in env')
-  } else if ( ! proxyUrl.includes(":") ) {
-    proxyUrl = proxyUrl + ":" + 9998
+  let proxyUrl = process.env.proxyServerIP;
+  if (!proxyUrl) {
+    throw Error('ERROR: Proxy Url not found in env');
+  } else if (!proxyUrl.includes(':')) {
+    proxyUrl = proxyUrl + ':' + 9998;
```
kschrief, line 120: I know this wasn't a part of the original ticket but let's just throw a log here saying we're using the default 9998 port

kschrief (changes requested):

server/src/proxyManagement.mjs
ksentak: Reverted to use previous log message

server/src/proxyManagement.mjs
ksentak: Agreed that this is unnecessary code so I removed line 105

server/src/proxyManagement.mjs
ksentak: Add additional console log

kschrief (dismissed):

server/src/messageHandler.mjs
```diff
@@ -285,7 +280,7 @@ async function handleMessage(message, userId, ws) {
 
   logger.debug(`Sending response for method ${oMsg.method}`);
   let finalResponse = (newResponse ? newResponse : response);
-  if (!process.env.proxy) {
+  if (!process.env.proxy || stateManagement.hasOverride(userId, oMsg.method)) {
```
michaelcfine, line 283: I'm not 100% sure I'm following the code changes here, but doesn't this change make it so that if there's an override, there won't be a response message sent back?!?

server/src/proxyManagement.mjs
```diff
@@ -21,97 +21,114 @@
...
 
-let ws = null
-let responseObject = {}
+const wsMap = new Map();
+const wsMsgMap = new Map();
 
-async function initialize(callback, returnWS) {
-  if(ws) {
-    return true
+async function initializeAndSendRequest(returnWs, command) {
```
michaelcfine, line 27: It seems a bit odd to combine 'initialize' and 'sendRequest'. Like... if you're going to sendRequest multiple times, it doesn't seem right that you also want to initialize each time. If the initialize is really "initialize if necessary" then maybe just don't mention it in the name of the function.

michaelcfine (commented):
I don't really understand the nature of the changes here. See one comment about the naming of one of the methods and a few minor comments. If you're sure this all works, I can approve without really understanding fully what you're changing / doing here.

ksentak:
Hey @michaelcfine, I want to give some more background regarding this PR so you can better understand the code changes:

Currently, on each new incoming connection to Mock Firebolt in Proxy Mode a WS object is created and assigned to the user. That same WS object is used to create a "messageHandler" for that user. Inside of that message handler, if Proxy Mode is enabled, a message is forwarded to the proxied endpoint. That forwarding is done through a separate WS object.

That proxying is currently done through a singleton WS object. Ripple opens several connections to MF. The result is mapping several incoming WS objects to one outgoing WS object. An unintended consequence of this is that when events/replies come in from the proxied endpoint, they are not going to the correct incoming WS object and events are not working properly on a Ripple connection.

A solution to the issue mentioned in the above paragraph is to turn the outgoing WS singleton into a 1:1 mapping with each incoming WS object.

When an incoming WS requests a message sent to the proxied endpoint, we now:

Check a WS:WS map to see if the incoming WS object has an outgoing WS object
If the map contains an entry for the incoming WS object, use it
If the map does not contain an entry, create an outgoing WS object and assign it to the map

server/src/messageHandler.mjs
kschrief: This was my addition.

I noticed this when testing setting overrides while in proxy mode. We haven't actually done that yet.

This block is saying that if we're not proxying, format a response based on whatever we got from the openRpc/method override. Otherwise if we're in proxy mode just blindly send back the message we got from the endpoint.

If we have an override for a method we need to treat it the same way we would for an openRPC. We need to format the canned response with the typical JsonRPC overhead and then send it.

I find it easier to think of this if/else in reverse: If we're returning a message we got from the thunder proxy endpoint, blindly send it as is. Otherwise we got some kind of mocked message (OpenRPC, method override, etc) that does not have the jsonRpc overhead. Add said overhead and return it.

server/src/proxyManagement.mjs
ksentak: You are correct in stating initialize is really an "initialize if necessary". Per your request, I have updated the function to be named `sendRequest()`.

michaelcfine (approved):
👍 despite not fully understanding the changes / intent here. counting on you / others to fully test.

kschrief (approved):

PR #119, opened 2023-04-04
Code changes for Parsing userID

kdivya153:


server/src/stateManagement.mjs
```diff
@@ -86,61 +86,13 @@ function addUser(userId) {
     logger.info(`Cannot add user ${userId}, already exists`);
     return {isSuccess: false, msg : `Cannot add user, already exists`};
   }
-
+  //To check whether the user,appId already exist 
+  let userExist = isUserExist(users,userId)
+  if((!userExist.isSuccess)){
```
kschrief, line 91: 1) Extra brackets
2) Let's also check to be sure userExist is there as well. This could throw an exception.

server/src/stateManagement.mjs
```diff
@@ -731,6 +683,50 @@ function deleteScratch(userId, key, scope=""){
...
 }
 
+/* 
+* @function:isUserExist()
+* @Description: check whether the user/appid/group is already exist
+* @Params:userId,users
+* @Return: JSON object {isSuccess: true/false, msg : ``}
+*/
+
+function isUserExist(users,userId){
```
kschrief, line 693: Let's rename to "doesUserExist" just to make it flow a little better in the code.

server/src/stateManagement.mjs
```diff
@@ -731,6 +683,50 @@ function deleteScratch(userId, key, scope=""){
   }
 }
 
+/* 
+* @function:isUserExist()
+* @Description: check whether the user/appid/group is already exist
```
kschrief: Quick grammar correction: "Check if the user/appId/group exists"

server/src/userManagement.mjs
```diff
@@ -174,61 +174,12 @@ function heartbeat(ws) {
...
-    }
-    else{
-      if (user && users[key] == user){
-        logger.info(`Cannot map user ${userId} to ws as user ${user} already exists`)
-        return false
-      }
-    }
+  //To check whether the user is already exist
+  let userExist =isUserExist(users,userId)
+  if(!userExist.isSuccess){
```
kschrief, line 179: Let's check to be sure userExist is there as well. This could throw an exception.

kschrief (changes requested):

server/src/stateManagement.mjs
kdivya153: Addressed

server/src/stateManagement.mjs
kdivya153: Addressed

server/src/userManagement.mjs
kdivya153: Addressed

kschrief (approved):

michaelcfine (changes requested):
The second set of changes looks great, fully adopting the utility function. But the code marked with my comment looks like it can be improved by using the utility function a second time within the loop.

server/src/stateManagement.mjs
```diff
@@ -697,22 +697,25 @@ function doesUserExist(users,userId){
 
   //iterating over list of users in state to ensure duplicate user/appId
   for(var key in users){
+    let parsedExistingUserId = parseUser(users[key])
+    let existingUser = parsedExistingUserId.user;
+    let existingAppId = parsedExistingUserId.appId;
     if (users[key].includes("~")){
-      if (user && users[key].split("~")[0]==user){
+      if (user && existingUser==user){
```
michaelcfine, line 704: This line is comparing two objects with equality, which is probably not what you want. Your variable names make distinguishing what's an object and what's an atomic field difficult. You might benefit from object dereferencing syntax (e.g., `const { existingUser, existingAppId, existingGroupId } = parseUser(users[key]);`).

server/src/stateManagement.mjs
```diff
@@ -697,22 +697,25 @@ function doesUserExist(users,userId){
...
         logger.info(`Cannot add user ${userId} as user ${user} already exists`)
         return {isSuccess: false, msg : `Cannot add user ${userId} as user ${user} already exists`}
       }
-      else if(appId && users[key].includes("#") && users[key].split("#")[1]==appId){
+      else if(appId && users[key].includes("#") && existingAppId==appId){
```
michaelcfine, line 708: If you use parseUser on both, I don't think you want to check for '#'... parsing's already done. You might want something like `else if (existingAppId === appId)` or perhaps `else if ( existingAppId && existingAppId === appId )` if you need to confirm that they are non-null.

michaelcfine (changes requested):
Still some work to do...

server/src/stateManagement.mjs
michaelcfine: Actually, I'm realizing now that these aren't objects. But I stand by my comment: either use object dereferencing so as to not even have variable names for the structs/objects or name them so that they are obviously objects, e.g., objUser, objExistingUser or similar.

scthunderbolt:

PR #121, opened 2023-04-06
Changes for allowing Mixed Case

kdivya153:


kschrief:
PR No longer needed. Change not appropriate for use cases.

PR #122, opened 2023-04-13
SYSTEST-9338 - Send events to the WS object that subscribed to it

ksentak:
**Background**
Mock Firebolt now supports multiple connections per user by utilizing an array/map of WS connections per userId.

This presents a problem for situations in which MF is tasked with triggering an event for a user. If a user is connected to MF with multiple connections, there is currently no way for MF to know which connection to send the requested event to.

In order to properly inject events, we must know which connection began listening to the requested event.

**Implementation**
MF already keeps track of which users are listening to which events. This is in order to support "sendEvent()" and "sendBroadcastEvent()" functionalities in YAML overrides as well as sending events manually using the --event CLI and its associated API. This same functionality can be extended to include which connection is listening to an event.

From there, the event functionality can be enhanced to send events to only the connections that are listening for it.

**Changes Made**
* Pass WS object to _registerEventListener_ and _deregisterEventListener_ functions inside of _messageHandler.mjs_.
* Modify _registerEventListener_, _deregisterEventListener_, & _emitResponse_ functions inside of _events.mjs_.
* _registerEventHandler_ now takes in a ws object and it is added in the eventListenerMap. 
* When it is time to call _emitResponse_, the WS object is retrieved from the eventListenerMap and used to send the response.

### DIFF
--- a/server/src/events.mjs
+++ b/server/src/events.mjs
@@ -58,7 +58,11 @@ function registerEventListener(userId, oMsg, ws) {
     eventListenerMap[userId][oMsg.method] = { id: oMsg.id, wsArr: [] };
   }
 
-  eventListenerMap[userId][oMsg.method].wsArr.push(ws);
+  // Check if ws is already in the wsArr before pushing
+  if (!eventListenerMap[userId][oMsg.method].wsArr.includes(ws)) {
+    eventListenerMap[userId][oMsg.method].wsArr.push(ws);
+  }
+
   logger.debug(`Registered event listener mapping: ${userId}:${oMsg.method}:${oMsg.id}`);
 }
 
--- a/server/src/events.mjs
+++ b/server/src/events.mjs
@@ -50,10 +50,15 @@ const eventListenerMap = {};
 
 // Associate this message ID with this method so if/when events are sent, we know which message ID to use
 function registerEventListener(userId, oMsg, ws) {
-  if ( ! eventListenerMap[userId] ) {
+  if (!eventListenerMap[userId]) {
     eventListenerMap[userId] = {};
   }
-  eventListenerMap[userId][oMsg.method] = { id: oMsg.id, ws: ws };
+
+  if (!eventListenerMap[userId][oMsg.method]) {
+    eventListenerMap[userId][oMsg.method] = { id: oMsg.id, wsArr: [] };
+  }
+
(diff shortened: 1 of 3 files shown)


kschrief (dismissed):

server/src/events.mjs
```diff
@@ -80,12 +80,14 @@ function getRegisteredEventListener(userId, method) {
...
 // Remove mapping from event listener request method name from our map
 // Attempts to send events to this listener going forward will fail
-function deregisterEventListener(userId, oMsg) {
-  if ( ! eventListenerMap[userId] ) { return; }
-  delete eventListenerMap[userId][oMsg.method];
-  logger.debug(`Deregistered event listener for method: ${userId}:${oMsg.method}`);
+function deregisterEventListener(userId, oMsg, ws) {
+  if (!eventListenerMap[userId]) { return; }
+  
+  if ( eventListenerMap[userId] && eventListenerMap[userId][oMsg.method] && eventListenerMap[userId][oMsg.method].ws === ws ) {
```
michaelcfine, line 86: Should we log some kind of error if the ws values don't match? Would this be an unexpected error to somehow track down? Or... should we *not* compare ws values and nuke the listener record no matter what?

michaelcfine (approved):
Nice work! One question/comment that I don't really know the answer. Depending on your confidence with how it is now, feel free to merge. Or, if my question leads you to consider a change, just ping me again when changed.

server/src/events.mjs
kschrief: Hmm. That brings up a scenario I don't know if we thought of. What if a user is listening to the same method across multiple connections. Why they would do this I don't know but I guess it's possible.

Considering the fact that we're still limiting to one listen per user per method I'd say we just nuke it no matter what. But maybe we want to put in the feature for multiple listens per user per connection?

Keaton, what'd be the difficulty on that? I'm assuming we'd just be setting [ws] as an array and some logic on top to add/remove the proper one.

ksentak: Per your request, I was able to set ws key to an array. I have renamed it to wsArr and we can now store multiple ws connections inside of the array.

michaelcfine (dismissed):
👍 good catches/fixes/improvements

kschrief (approved):

michaelcfine (approved):
Ah, good catch! Didn't think of that.

PR #123, opened 2023-04-14
SYSTEST-9348-Use the word "group" as a shortcut for the user's group in ctx functions

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/SYSTEST-9348
As part of this ticket, implemented changes to support
- Allowing a "group" keyword to be passed for "scope" for all ctx.set(), ctx.get() and ctx.delete() functions
- When passed, the group of the userId that is INVOKING the YAML will be used as the "scope" for that call
- If the userId does not contain a group, a warning should be logged and full userId is used instead

### DIFF
--- a/server/src/stateManagement.mjs
+++ b/server/src/stateManagement.mjs
@@ -705,19 +705,19 @@ and if userid invoking the YAML contains group, the group of the userId will be 
 else,  the full userid is used along with warning
 */
 function setScopeForGroupKeyword(userId, scope) {
-  let parseUserId = parseUser(userId)
-  if (parseUserId.group == undefined || parseUserId.group == null) {
+  let parsedUserId = parseUser(userId)
+  if (!parsedUserId.group) {
     logger.warning("WARNING: userId does not contain a group. Using full userId for manipulating scratch")
     scope = userId
   } else {
-    scope = '~' + parseUserId.group
+    scope = '~' + parsedUserId.group
   }
   return scope
 }
 
 // set scratch space of scope with the provided key-value
 function setScratch(userId, key, val, scope) {
-  if (scope == 'group') {
+  if (scope === 'group') {
     scope = setScopeForGroupKeyword(userId, scope)
   }
   updateState(userId, {
@@ -729,7 +729,7 @@ function setScratch(userId, key, val, scope) {
 
 // get key from scratch space of provided scope
 function getScratch(userId, key, scope) {
-  if (scope == 'group') {
+  if (scope === 'group') {
     userId = setScopeForGroupKeyword(userId, scope)
   }
   const userState = getState(userId);
@@ -742,7 +742,7 @@ function getScratch(userId, key, scope) {
 // delete key from scratch space of provided scope
 function deleteScratch(userId, key, scope=""){
   if (scope !== "") {
-    if (scope == 'group') {
+    if (scope === 'group') {
       scope = setScopeForGroupKeyword(userId, scope)
     }
     if (scope in state && key in state[scope].scratch) {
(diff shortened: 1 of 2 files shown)


server/src/stateManagement.mjs
```diff
@@ -700,17 +700,39 @@ function setMethodError(userId, methodName, code, message) {
...
 }
 
+/* if "group" keyword is passed for scope inside ctx functions, 
+and if userid invoking the YAML contains group, the group of the userId will be used as the "scope" for that call
+else,  the full userid is used along with warning
+*/
+function setScopeForGroupKeyword(userId, scope) {
+  let parseUserId = parseUser(userId)
+  if (parseUserId.group == undefined || parseUserId.group == null) {
+    logger.warning("UserId does not contain a group. Using full userid for manipulating scratch")
```
kschrief, line 710: Inconsistent capitalization. We should use a capital "I" in "userId"

kschrief (changes requested):

server/src/stateManagement.mjs
```diff
@@ -700,17 +700,39 @@ function setMethodError(userId, methodName, code, message) {
...
+and if userid invoking the YAML contains group, the group of the userId will be used as the "scope" for that call
+else,  the full userid is used along with warning
+*/
+function setScopeForGroupKeyword(userId, scope) {
+  let parseUserId = parseUser(userId)
+  if (parseUserId.group == undefined || parseUserId.group == null) {
+    logger.warning("WARNING: userId does not contain a group. Using full userId for manipulating scratch")
+    scope = userId
+  } else {
+    scope = '~' + parseUserId.group
```
michaelcfine, line 713: I don't know where else "special string construction or parsing" is done for full userId values, but it might be nice to create some kind of "createUserIdForGroup" function for this (alongside the parseUser utility function).

server/src/stateManagement.mjs
```diff
@@ -700,17 +700,39 @@ function setMethodError(userId, methodName, code, message) {
   });
 }
 
+/* if "group" keyword is passed for scope inside ctx functions, 
+and if userid invoking the YAML contains group, the group of the userId will be used as the "scope" for that call
+else,  the full userid is used along with warning
+*/
+function setScopeForGroupKeyword(userId, scope) {
+  let parseUserId = parseUser(userId)
+  if (parseUserId.group == undefined || parseUserId.group == null) {
```
michaelcfine, line 709: minor, but in JS, you can just test for falsiness... "if ( ! parseUserId.group ) {"
also minor, but probably "parsedUserId" (note the 'd')

server/src/stateManagement.mjs
```diff
@@ -700,17 +700,39 @@ function setMethodError(userId, methodName, code, message) {
...
+    scope = userId
+  } else {
+    scope = '~' + parseUserId.group
+  }
+  return scope
+}
+
+// set scratch space of scope with the provided key-value
 function setScratch(userId, key, val, scope) {
+  if (scope == 'group') {
```
michaelcfine, line 720: minor: generally, use === in JS instead of ==.

michaelcfine (approved):
A few minor comments but otherwise I think this is fine. I'm still quite surprised / confused by all the code related to scope, so I'm counting on you / others that this stuff is all necessary.

Nandana-NNR (commented):
@michaelcfine
Just to clarify, currently a user must explicitly give a group name in ctx function of yaml. This limits the YAMLs usage to just that group (Ex: set(key, value, "~A) making that YAML only usable for users in group ~A). Now, with the 'group' keyword, user can set the scratch variable for an entire group without defining that group in the YAML manually
Also observing that in yesterday's merge from core-sdk team(https://github.com/rdkcentral/firebolt-core-sdk/commit/1a662eed1e5f1ef00d57187ccf883ced7b02240c), 
sdk version has now been updated and the new openrpc has method starting in capital letter, whereas the current  method in functional test starts with small letter, leading to functional test failure. Hence, updated the same in functional test in the PR
cc @kschrief

PR #124, opened 2023-04-18
parse the userId

kdivya153:


kschrief (approved):

michaelcfine (approved):
👍

server/src/stateManagement.mjs
```diff
@@ -160,6 +137,7 @@ function getUserId(userId){
   // //Checking if user or appId is present in state object
   if (!(userId in state)){
     for(var key in users){
+
```
michaelcfine, line 140: I still see uses of "~", which seems less than ideal. Once the currentUserId is parsed and each userId is parsed within the loop, I don't think there should be any string-related manipulation going on.

michaelcfine (changes requested):
The parsedUserId change is good, but I still see uses of "~", etc. in the code. Once both userid values are parsed (one at the top, one in the loop), there shouldn't be a need for string inspection anymore.

server/src/stateManagement.mjs
kdivya153: Thanks Mike for pointing it out,We have separate ticket to handle the user inside the loop.This ticket is only for ParsedUserId().Here is the ticket:https://ccp.sys.comcast.net/browse/SYSTEST-9362

michaelcfine (approved):
👍 due to ticket confusion

kschrief (approved):

PR #126, opened 2023-04-27
Support any JsonRPC structure for events - Part 1

ksentak:
MF has the ability to trigger events either manually through the API/CLI or automatically through triggers.

In order to send and format these events, MF must keep track of which incoming users are listening. This formatting/searching/etc is currently hard-coded for the FB format. We must open up the existing event functionality to support any JsonRPC structure for events.

We need to convert the existing hard-coded logic into something that's configurable. This is part 1 of this implementation and includes implementation of the "registrationMessage" and "unRegistrationMessage" configuration elements.

Message Metadata
When a new event registration/unregistration is detected, the message should be saved as additional metadata for the user/method being registered/unregistered. This will be used later on to format ack messages using Handlebars.js.

The registration message object should be saved as "registration"
The event message object should be saved as "event"
The unRegistration message object should be saved as "unRegistration"

### DIFF
--- a/server/src/events.mjs
+++ b/server/src/events.mjs
@@ -20,6 +20,7 @@
 
 'use strict';
 
+import JSONPath from 'jsonpath';
 import * as stateManagement from './stateManagement.mjs';
 import * as userManagement from './userManagement.mjs';
 import { eventTriggers } from './triggers.mjs';
@@ -27,6 +28,8 @@ import { logger } from './logger.mjs';
 import * as fireboltOpenRpc from './fireboltOpenRpc.mjs';
 import { config } from './config.mjs';
 import { updateCallWithResponse } from './sessionManagement.mjs';
+
+const { dotConfig: { eventConfig } } = config;
 
 function logSuccess(onMethod, result, msg) {
   logger.info(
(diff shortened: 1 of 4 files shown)


server/src/messageHandler.mjs
```diff
@@ -80,17 +83,17 @@ async function handleMessage(message, userId, ws) {
   }
 
   // Handle JSON-RPC messages that are event listener enable requests
-
   if ( events.isEventListenerOnMessage(oMsg) ) {
     events.sendEventListenerAck(userId, ws, oMsg);
-    events.registerEventListener(userId, oMsg, ws);
+    const eventMetadata = events.extractEventData(oMsg, eventConfig.registrationMessage, true);
```
kschrief, line 88: Let's leave a comment here on what this line's doing.

server/src/messageHandler.mjs
```diff
@@ -80,17 +83,17 @@ async function handleMessage(message, userId, ws) {
...
+    events.registerEventListener(userId, eventMetadata, ws);
     return;
   }
-
+  
   // Handle JSON-RPC messages that are event listener disable requests
-
   if ( events.isEventListenerOffMessage(oMsg) ) {
-    events.deregisterEventListener(userId, oMsg, ws);
+    const eventMetadata = events.extractEventData(oMsg, eventConfig.unRegistrationMessage, false);
```
kschrief, line 95: Let's leave a comment here on what this line's doing.

server/src/events.mjs
```diff
@@ -87,48 +97,112 @@ function getRegisteredEventListener(userId, method) {
...
+ * Determines if the given object message is an event listener message.
+ * @param {object} oMsg - The object message to check.
+ * @param {object} config - The configuration object to use.
+ * @returns {boolean} - True if the message is an event listener message, false otherwise.
+*/
+function isEventListenerMessage(oMsg, config) {
+  // Call extractEventData and store the result
+  const eventData = extractEventData(oMsg, config);
+
+  if (eventData !== false) {
```
kschrief, line 175: return (eventData !== false)

:)

server/src/events.mjs
```diff
@@ -87,48 +97,112 @@ function getRegisteredEventListener(userId, method) {
...
+  const extractedMethod = JSONPath.query(oMsg, method);
+
+  if (extractedMethod.length === 0) {
+    return false;
+  }
+
+  const methodName = extractedMethod[0];
+  const metadata = {
+    registration: {},
+    event: methodName,
```
kschrief, line 150: This goes against the design.

metadata.event should contain the entire event payload that's being processed. We should be able to retrieve the method being followed through eventListenerMap[userId][method]

kschrief: Note that I'm on-board with adding a "method" field to the metadata and putting it there if it makes life easier.

kschrief (changes requested):

server/src/events.mjs
```diff
@@ -87,48 +99,108 @@ function getRegisteredEventListener(userId, method) {
...
+
+/**
+ * Extracts event data from a given message object based on provided configuration.
+ * @param {object} oMsg - The message object to extract event data from.
+ * @param {object} config - The configuration object for the event data extraction.
+ * @param {boolean} isOn - Whether eventListener enable or disable request
+ * @returns {object | false} - An object containing the extracted event data or false if the extraction fails.
+*/
+function extractEventData(oMsg, config, isOn) {
+  const { searchRegex, method } = config;
```
kschrief, line 137: What happens to extractedMethod if the configured method is null/empty? It's optional and we should default to "$.method".

kschrief (changes requested):

PR #130, opened 2023-05-10
SYSTEST-9379-Remove branding references from MF

Nandana-NNR:
In several places in Mock Firebolt (Documentation, tests etc) there are references to real-world applications that might be used in MF (Ex: "netflix", "youtube"). As part of this ticket, 

- Removed references to non-Comcast applications inside of any Mock Firebolt code.

- Replaced such references with generic names
Ticket - https://ccp.sys.comcast.net/browse/SYSTEST-9379

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -70,7 +70,7 @@ run `node cli.mjs --getStatus`
 
 To add user :
 run `node cli.mjs --addUser <userId>`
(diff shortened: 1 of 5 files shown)


kschrief (approved):

michaelcfine (approved):
👍

PR #131, opened 2023-05-12
SYSTEST-9335-Remove manage functionality

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/SYSTEST-9427

This ticket is to retrieve the MonoSDK from HTTP and remove the "Manage" and "Core" dependencies from Mock-Firebolt and covers the below points -
1) The "manage" dependency is removed from MF as well as anything hard-coded to work with it
2) The "core" dependency is moved from a dependency to a configuration referring to the new online source for the mono-repo.
3) The default configuration is updated to load the mono-repo OpenRPC from the online source and enable 'allowMixedCase' flag to true
4) The MF code is updated to handle self-referencing OpenRPC documents
5) The "discovery" dependency is not altered


README.md
```diff
@@ -136,28 +134,17 @@ cd server
 
 cp src/.mf.config.SAMPLE.json src/.mf.config.json
 
+# To install dependencies, clean/create build/ subdirectory, build and upgrade SDK, build source code within this project
 npm install
-npm run clean             # Cleans/creates build/ subdirectory
```
kschrief, line 140: This is still required for discovery.

README.md
```diff
@@ -136,28 +134,17 @@ cd server
...
 
-npm run build:core        # MANDATORY: If you're only using the Firebolt Core SDK (typical for most 3rd party app devs)
-npm run build:manage      # OPTIONAL: If you're using the Firebolt Manage SDK (not typical)
-npm run build:discovery   # OPTIONAL: If you're using the Firebolt Discovery SDK (not typical)
```
kschrief, line 144: We are only removing discovery from testing. All regular functionality regarding discovery should stay.

README.md
```diff
@@ -136,28 +134,17 @@ cd server
...
-
-npm run build:mf          # Builds source code within this project
```
kschrief, line 146: This is not dependent on the SDKs and should stay.

Dockerfile.allSdks
```diff
@@ -48,5 +45,5 @@ ENV PATH=${PATH}:/usr/src/firebolt/mock-firebolt/cli
 WORKDIR /usr/src/firebolt/mock-firebolt/server
 
 ENTRYPOINT [ "npm",  "run", "start" ]
-# By default, only the core SDK is enabled. When using 'docker run', users can pass --manage --discovery as command-line parameters
+# By default, only the mono-sdk retrieved from HTTP is enabled. When using 'docker run', users can pass --discovery as command-line parameters
```
kschrief, line 48: I think the mono-sdk is still just being referred to as "core". Our documentation should match that. Can we confirm what the mono SDK is being called?

server/src/fireboltOpenRpc.mjs
```diff
@@ -284,7 +284,7 @@ async function readSdkJsonFileIfEnabled(sdkName) {
         rawMeta[sdkName] = JSON.parse(
           await readFile(fileUrl)
         );
-        logger.info(`Loaded ${sdkName} SDK from URL ${url}`);
+        logger.info(`Loaded mono SDK from URL ${url}`);
```
kschrief, line 287: This shouldn't change. We can load more than just the mono SDK in this way.

kschrief (changes requested):

Dockerfile.allSdks
Nandana-NNR: Checking with Jeremy on the same

README.md
Nandana-NNR:
```
   "upgradeSDKAndBuildMF": "npm update @firebolt-js/discovery-sdk && npm run clean && npm run build:discovery && npm run build:mf",
    "postinstall": "npm run upgradeSDKAndBuildMF"
```
I have only removed the core and manage steps, all the remaining steps(related to discovery,clean and create build folder, building src code) are moved into a postinstall script in server/package.json as above. Thought this would be better and  easy on the user as user now only has to run 'npm install' manually instead of each of the steps and rest will be taken care by the same command in the background

README.md
Nandana-NNR: Same as above comment

README.md
Nandana-NNR: Same as initial comment

Dockerfile.coreSdk
```diff
@@ -46,5 +46,5 @@ ENV PATH=${PATH}:/usr/src/firebolt/mock-firebolt/cli
 WORKDIR /usr/src/firebolt/mock-firebolt/server
```
michaelcfine, line 47: Having both ".coreSdk" and ".allSdks" Dockerfiles is probably unnecessary now that things are merged together and you can't really have one SDK without the other. I think we should merge these / just have a "Dockerfile".

README.md
```diff
@@ -25,7 +25,7 @@ Note, too, that since the SDK does not validate parameters on method calls, app
 - Controllable Mock Firebolt server (NodeJS websocket + HTTP server)
 - Control mock method responses via control plane RESTful API, CLI, web admin UI (soon), browser extension (soon)
 - 100% OpenRPC-driven; no SDK-specific details within the implementation
-- Supports Firebolt SDKs: **core** (always), **manage** (optional), **discovery** (optional)
+- Supports Firebolt SDKs: **core** (always), **discovery** (optional)
```
michaelcfine, line 28: I still think we should say we support the manage SDK, since we do. It's just that the OpenRPC for it is included in the one unified OpenRPC file.

If you don't want to mention the 'manage' stuff, then I think instead of 'SDK' you have to say 'OpenRPC' or something like that, not SDK.

README.md
```diff
@@ -87,13 +87,13 @@ The next two sections presume you are using the out-of-the-box `.mf.config.json`
 
 # Firebolt SDK Support
 
-Mock Firebolt supports these Firebolt SDKs: **core**, **manage**, **discovery**.
+Mock Firebolt supports these Firebolt SDKs: **core**, **discovery**.
```
michaelcfine, line 90: See earlier comment. From an SDK perspective, we still do support the manage SDK.

README.md
```diff
@@ -87,13 +87,13 @@ The next two sections presume you are using the out-of-the-box `.mf.config.json`
...
 
-By default, only the core SDK is enabled, meaning if you try to make calls to methods in the other SDKs, Mock Firebolt will thow an error.
+By default, only the core sdk retrieved from HTTP is enabled
```
michaelcfine, line 92: Again, I don't think this is quite right... Only the unified OpenRPC retrieved via HTTP is used, but this does mean that both/all SDKs are 'enabled' (supported).

README.md
```diff
@@ -109,13 +109,11 @@ Developers wishing to activate this functionality must pass the `--moneybadger`
 
 | SDK                                 | Minimum Version |
 | ----------------------------------- | --------------- |
-| @firebolt-js/firebolt-sdk           | 0.6.0-alpha.1   |
-| @firebolt-js/firebolt-manage-sdk    | 0.1.0-alpha.2   |
 | @firebolt-js/firebolt-discovery-sdk | 0.1.0-alpha.1   |
```
michaelcfine, line 112: Technically, the deleted comments about minimum versions of the SDKs are correct. But those are old enough now that I think we should just remove this section about minimum versions.

README.md
```diff
@@ -184,9 +171,9 @@ Click the "Load Unpacked" button (top left corner of browser window)
 Navigate to the directory under browser-extensions which contains a manifest.json file and click "Select"
 ```
 
-Now you can add `@firebolt-js/sdk` in your app's `package.json` file, import `@firebolt-js/sdk` in your code, and make calls using the Firebolt core SDK (or use the manage and/or discovery SDKs too). If you pass the appropriate query string parameter(s) (see [docs/UsageWithinApps.md](./docs/UsageWithinApps.md)), these SDK calls will get sent to the Mock Firebolt server and it will reply as you configure it to do so.
+Now you can access core sdk from HTTP by adding `"url": "https://meta.rdkcentral.com/firebolt/api/"` in mf.config.SAMPLE.json and copying to mf.config.json,If you pass the appropriate query string parameter(s) (see [docs/UsageWithinApps.md](./docs/UsageWithinApps.md)), these SDK calls will get sent to the Mock Firebolt server and it will reply as you configure it to do so.
 
-See the section "Important Dependency Notes" above for details about which versions of `@firebolt-js/sdk` (and other SDKs) support Mock Firebolt.
+See the section "Important Dependency Notes" above for details about which version of SDK support Mock Firebolt.
```
michaelcfine, line 176: If we delete that section (see earlier comment), we should delete this note.

README.md
```diff
@@ -205,7 +192,7 @@ docker build -f Dockerfile.allSdks . -t $MF_DOCKER_USER/mock-firebolt
 # Confirm your image was created
 docker images | grep mock-firebolt
 
-# Run the image, enabling just the core SDK (typical)
+# Run the image, enabling the core SDK (typical)
```
michaelcfine, line 195: Again, technically, we are enabling the manage SDK since the unified OpenRPC has the methods/messages it supports.

docs/DotConfig.md
```diff
@@ -2,20 +2,20 @@ Mock Firebolt: mf.config.json(Supported SDK's)
 =======================
 ## Supported SDK's
 
-Mock Firebolt supports these Firebolt SDKs: core, manage, discovery.
-mf.config.json file contains a list of SDK's supported by Mock Firebolt, Out of which `core` is mandatory and other two `manage` and `discovery` are optional.
+Mock Firebolt supports these Firebolt SDKs: core, discovery.
```
michaelcfine, line 5: We do support the manage SDK since the unified file includes its methods/messages.

docs/DotConfig.md
```diff
@@ -2,20 +2,20 @@ Mock Firebolt: mf.config.json(Supported SDK's)
...
+mf.config.json file contains a list of SDK's supported by Mock Firebolt, Out of which `core` is mandatory and `discovery` is optional.
```
michaelcfine, line 6: We should probably talk about this. The SDK-ness of this feature is hard to support now that the open rpc for both core and manage (and maybe discovery in the future) is all in one OpenRPC file.

docs/DotConfig.md
```diff
@@ -2,20 +2,20 @@ Mock Firebolt: mf.config.json(Supported SDK's)
...
 and it contains associated command-line flags you provide when you start the server.
-# If you're only using the Firebolt Core SDK (typical for most 3rd party app developers)
+# If you're using the Firebolt core SDK
```
michaelcfine, line 8: Not sure why you deleted the comment.

functional/tests/utilities.mjs
```diff
@@ -24,7 +24,7 @@ import path from "path";
 
 let mfProcess;
 let startMf = "npm run dev";
-const mfStarted = "Listening on HTTP port";
+const mfStartedAndSdkLoaded = "Loaded core SDK from URL";
```
michaelcfine, line 27: again, you're not loading the "core 'SDK'", you're loading any/all open rpc messages in the unified file, which happen to be spread across both SDKs.

server/package.json
```diff
@@ -6,21 +6,20 @@
...
     "build:discovery": "node ./scripts/copy.mjs ./node_modules/@firebolt-js/discovery-sdk/dist/firebolt-discovery-open-rpc.json ./build/firebolt-discovery-open-rpc.json",
     "build:openrpc:NOTES": "Only use if you are using all three Firebolt SDKs",
-    "build:openrpc": "npm run build:core && npm run build:manage && npm run build:discovery",
+    "build:openrpc": "npm run build:discovery",
     "build:mf": "node ./scripts/rcopy.mjs ./src ./build && node ./scripts/copy.mjs ./src/.mf.config.json ./build/.mf.config.json && node ./scripts/copy.mjs ./package.json ./build/package.json",
     "build:NOTES": "Only use if you are using all three Firebolt SDKs",
     "build": "npm run build:openrpc && npm run build:mf",
     "test": "NODE_OPTIONS=--experimental-vm-modules NODE_ENV=test npx jest --config=jest.config.js --detectOpenHandles --runInBand --forceExit",
-    "start:NOTES1": "Can do 'npm run start' (core SDK only) or 'npm run start --manage --discovery' (all three SDKs)",
+    "start:NOTES1": "Can do 'npm run start' (core SDK only) or 'npm run start -- --discovery'",
```
michaelcfine, line 16: again, not really "core SDK"...

server/src/.mf.config.SAMPLE.json
```diff
@@ -9,18 +9,10 @@
             "name": "core",
             "cliFlag": null,
             "cliShortFlag": null,
-            "fileName": "firebolt-open-rpc.json",
+            "url": "https://meta.rdkcentral.com/firebolt/api/",
```
michaelcfine, line 12: I think the fact that the parent key is "supportedSdks" is problematic now that we have one file that spans multiple SDKs. Maybe this should be something more like "supportedOpenRpcDefinitions" or "supportedInterfaces" or something like that. Or maybe we remove this feature completely somehow. Again, the problem is that there's now no longer a 1-1 mapping between SDKs and OpenRPC files.

server/src/commandLine.mjs
```diff
@@ -26,15 +26,14 @@ import { config } from './config.mjs';
 import { mergeArrayOfStrings } from './util.mjs';
 
 // Usage:
-//   node index.mjs                                          (core SDK only, default)
+//   node index.mjs                                          (core SDK, default)
```
michaelcfine, line 29: not really "core SDK"... "main FB OpenRPC file", yes; but still both/all SDKs.

michaelcfine (changes requested):
Lots of comments. Mainly: There is now no longer a 1-1 mapping between SDKs and OpenRPC files. In many cases, your changes talk about "the core SDK" but that's not right... by virtue of using the unified OpenRPC file, we do actually continue to support the manage SDK (and any others whose methods are in that OpenRPC file).

README.md
kschrief: Maybe we can change manage to "(always)" since it's loaded with core now.

README.md
kschrief: Agreed. We can say "only the core/manage SDK" or similar.

Dockerfile.allSdks
Nandana-NNR: confirmed the correct reference is core and updated the same

Dockerfile.coreSdk
Nandana-NNR: Addressed

README.md
Nandana-NNR: Addressed

README.md
Nandana-NNR: Addressed

README.md
Nandana-NNR: Addressed

README.md
Nandana-NNR: Addressed

docs/DotConfig.md
Nandana-NNR: Addressed

README.md
Nandana-NNR: Addressed

docs/DotConfig.md
Nandana-NNR: Updated

functional/tests/utilities.mjs
Nandana-NNR: This is because here we are waiting for the log message that gets generated after loading unified openrpc in fireboltOpenRpc.mjs .i.e:
`logger.info(Loaded ${sdkName} SDK from URL ${url});
`sdkName here refers to the name inside supportedSdks in config file, which in this case is 'core' which has the unifiedopenrpc url as manage is removed from supportedSdks
```
` `"supportedSdks": [

        {
            "name": "core",
            "cliFlag": null,
            "cliShortFlag": null,
            "url": "https://meta.rdkcentral.com/firebolt/api/",
            "enabled": true
        },
```

functional/tests/core-manage-discovery-jsonrpc.test.mjs
kschrief, line 1: Since "Discovery" doesn't exist anymore and we're using a dummy openRPC maybe we can just combine these into one file? Call it something like openRpcCall.test.mjs or similar?

functional/tests/core-discovery-jsonrpc.test.mjs
kschrief, line 1: Since "Discovery" doesn't exist anymore and we're using a dummy openRPC maybe we can just combine these into one file with core? Call it something like openRpcCall.test.mjs or similar?

functional/tests/core-manage-jsonrpc.test.mjs
kschrief, line 1: Since "Discovery" doesn't exist anymore and we're using a dummy openRPC maybe we can just combine these into one file with core? Call it something like openRpcCall.test.mjs or similar?

functional/tests/core.test.mjs
kschrief, line 1: Since "Discovery" doesn't exist anymore and we're using a dummy openRPC maybe we can just combine these into one file with core? Call it something like openRpcCall.test.mjs or similar?

kschrief (changes requested):

PR #132, opened 2023-05-17
Support JsonRPC structure for events - Part II

ksentak:
MF has the ability to trigger events either manually through the API/CLI or automatically through triggers.

In order to send and format these events, MF must keep track of which incoming users are listening. This formatting/searching/etc is currently hard-coded for the FB format. We must open up the existing event functionality to support any JsonRPC structure for events.

### DIFF
--- a/docs/Events.md
+++ b/docs/Events.md
@@ -1,9 +1,12 @@
-Mock Firebolt: Events: Misc Notes
-=================================
+# Mock Firebolt: Events: Misc Notes
 
-Example event request, event ack, and event
+This document covers the flow of event registration, acknowledgement, and actual event messages in the Mock Firebolt framework. It also explains the configuration used to manage event messages.
+
+## Flow of Events
 
 ### Event Request
+
+An event request is a message from a client to register for an event. Below is an example of an event request message:
 
 ```
 {"jsonrpc":"2.0","method":"lifecycle.onInactive","params":{"listen":true},"id":1}
@@ -11,12 +14,29 @@ Example event request, event ack, and event
 
 ### Event Acknowledgement
 
+Once the server has successfully registered the client for the specified event, it optionally sends back an acknowledgement message. Here is an example:
+
 ```
 {"jsonrpc":"2.0","result":{"listening":true, "event":"lifecycle.onInactive"},"id":1}
 ```
 
 ### Event
 
+When the registered event occurs, the server sends an event message to the client. Here is an example of an event message:
+
 ```
 {"jsonrpc":"2.0","result":{"state":"inactive","previous":"foreground"}, "id":1}
 ```
+
+## Configuration
+The configuration used to manage events is found in the `.mf.config.json` file under the `eventConfig` key. Here is an explanation of each field:
+
+* registrationMessage: This defines the search pattern to identify registration messages and the method to be used. The search pattern is a regular expression. The method key is a json path.
+
+* unRegistrationMessage: Similar to registrationMessage, this defines the search pattern to identify unregistration messages and the method to be used.
+
+* registrationAck (Optional): This is the template for the acknowledgement message sent when a client successfully registers for an event. It is a Handlebars template string where metadata placeholders get replaced with actual values.
+
+* unRegistrationAck (Optional): This is the template for the acknowledgement message sent when a client successfully unregisters from an event. It is similar to registrationAck.
+
+* event (Optional): This is the template for the actual event message sent when the registered event occurs. It is a Handlebars template where placeholders get replaced with actual values.
--- a/docs/Events.md
+++ b/docs/Events.md
@@ -1,6 +1,6 @@
 # Mock Firebolt: Events: Misc Notes
 
-This document covers the flow of event registration, acknowledgement, and actual event messages in the Mock Firebolt framework. It also explains the configuration used to manage event messages.
+This document covers the flow of event registration, acknowledgement, and actual event messages in the Mock Firebolt framework. It also explains the configuration used to manage event messages and walks you through a typical flow using default configurations.
 
 ## Flow of Events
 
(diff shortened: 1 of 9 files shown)


server/src/events.mjs
```diff
@@ -87,70 +100,152 @@ function getRegisteredEventListener(userId, method) {
...
+}
+
+/**
+ * Extracts event data from a given message object based on provided configuration.
+ * @param {object} oMsg - The message object to extract event data from.
+ * @param {object} config - The configuration object for the event data extraction.
+ * @param {boolean} isOn - Whether eventListener enable or disable request
+ * @returns {object | false} - An object containing the extracted event data or false if the extraction fails.
+*/
+function extractEventData(oMsg, config, isOn) {
```
kschrief, line 137: I'd recommend renaming "isOn" to something a bit more intuitive like "isEnable"

server/src/events.mjs
```diff
@@ -87,70 +100,152 @@ function getRegisteredEventListener(userId, method) {
...
+  const searchRegex = config.searchRegex;
+  const method = config.method || '$.method';
+
+  if (!new RegExp(searchRegex).test(JSON.stringify(oMsg))) {
+    return false;
   }
+
+  const extractedMethod = JSONPath.query(oMsg, method);
+
+  if (extractedMethod.length === 0) {
```
kschrief, line 147: Can we add a log here to indicate why we're failing?

docs/Events.md
```diff
@@ -1,22 +1,42 @@
...
 
 ### Event
 
+When the registered event occurs, the server sends an event message to the client. Here is an example of an event message:
+
 ```
 {"jsonrpc":"2.0","result":{"state":"inactive","previous":"foreground"}, "id":1}
 ```
+
+## Configuration
```
kschrief, line 31: We should add an example flow here. The user will need to know what they can access via handlebars and how. Walking through a flow using the default config in .mf.config.json would do nicely.

server/src/events.mjs
```diff
@@ -87,70 +100,152 @@ function getRegisteredEventListener(userId, method) {
...
+/**
+ * Determines whether the given message is intended to disable an event listener.
+ * A regex pattern and JSON path are provided in config to check whether the event listener is off.
+ * @param {Object} oMsg - The incoming message to check.
+ * @return {boolean} - Returns `true` if the message is intended to disable an event listener, `false` otherwise.
+ */
 function isEventListenerOffMessage(oMsg) {
-  if ( ! isEventListenerMessage(oMsg) ) { return false; }
-  return ( oMsg.params && ! oMsg.params.listen );
+  if (!isEventListenerMessage(oMsg, eventConfig.unRegistrationMessage)) {
```
kschrief, line 203: This seems overly complex.

I'd say we can make this a one-liner (return isEventListenerMessage(...)) but that also seems overly complex considering we can just use that one-liner where we'd want to call "isEventListenerOffMessage()".

Can this be simplified?

server/src/events.mjs
```diff
@@ -161,20 +256,34 @@ function sendBroadcastEvent(ws, userId, method, result, msg, fSuccess, fErr, fFa
...
+  // Defines the data object that will be inputted into handlebars
+  const templateData = {
+    ...metadata,
+    result: finalResult,
+    resultAsJson: JSON.stringify(finalResult)
   };
-  const eventMessage = JSON.stringify(oEventMessage);
+
+  // If event template config does not exist, use default template which just returns the result
+  const defaultEventTemplate = `{"result":{{{resultAsJson}}}{{~"}}"~}}}`;
```
kschrief, line 283: We shouldn't do any formatting at all unless a template exists. Just send the exact event payload that was given to the API

server/src/messageHandler.mjs
```diff
@@ -80,18 +83,47 @@ async function handleMessage(message, userId, ws) {
...
+      if (eventConfig.registrationAck) {
+        events.sendEventListenerAck(userId, ws, eventMetadata);
+      }
+  
+      events.registerEventListener(userId, eventMetadata, ws);
+      return;
+    }
+    // If in proxy mode, just register the event listener without sending ack messages or returning
+    else {
+      events.registerEventListener(userId, eventMetadata, ws);
```
kschrief, line 103: There's some duplicate code here. We always have to register the listener so we can do that first without putting it in the if/else block.

server/src/messageHandler.mjs
```diff
@@ -80,18 +83,47 @@ async function handleMessage(message, userId, ws) {
...
+      if (eventConfig.unRegistrationAck) {
+        events.sendUnRegistrationAck(userId, ws, eventMetadata);
+      }
+      
+      events.deregisterEventListener(userId, eventMetadata, ws);
+      return;
+    }
+    // If in proxy mode, just deregister the event listener without sending ack messages or returning
+    else {
+      events.deregisterEventListener(userId, eventMetadata, ws);
```
kschrief, line 125: See above duplicate code comment

server/src/events.mjs
ksentak: Added following log

    logger.debug(`Error occurred while extracting event data: No method found in the provided message.`);

server/src/events.mjs
ksentak: Kept separate functions so code is easier to read/understand but simplified the functions into one liners

kschrief (changes requested):

docs/Events.md
```diff
@@ -1,22 +1,100 @@
...
+
+```
+// Handlebars config
+{\"result\":{{{resultAsJson}}},\"id\":{{registration.id}},\"jsonrpc\":\"2.0\"}
+
+// Actual result
+{"result": "NEW-DEVICE-NAME-1","id":7,"jsonrpc":"2.0"}
+```
+
+In the Handlebars templates, you can access any data from the event metadata using the dot notation. For instance, you can use {{metadata.params.id}} to access the id from the params object in the metadata.
```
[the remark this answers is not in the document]
ksentak: In the Handlebars templates, you can access data from the event registration, unregistration, and the event itself, which we collectively refer to as metadata. You can use dot notation to access specific pieces of data. For instance, you can use {{registration.id}} to access the id from the registration object in the metadata. Similarly, you can access information from unregistration and event metadata objects.


Changed to^^

kschrief (approved):

server/src/events.mjs
```diff
@@ -48,22 +52,31 @@ function logFatalErr() {
...
+  const { registration, method } = metadata;
+
   if (!eventListenerMap[userId]) {
     eventListenerMap[userId] = {};
   }
 
-  if (!eventListenerMap[userId][oMsg.method]) {
-    eventListenerMap[userId][oMsg.method] = { id: oMsg.id, wsArr: [] };
+  if (!eventListenerMap[userId][method]) {
+    // id: registration.id will be removed in part 2
```
michaelcfine, line 70: Wondering if this represents something left-over / to do since this is the PR for part 2, right?

michaelcfine (approved):
👍 Looks great. One comment about a comment in the code that looks like maaaybe something didn't get done?.

server/src/events.mjs
ksentak: Nice catch, forgot to clean up!

kschrief (approved):

PR #133, opened 2023-06-07
SYSTEST-9427-Retrieve Mono-Repo from HTTP and remove Manage and Core dependencies

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/SYSTEST-9427

This ticket is to retrieve the MonoSDK from HTTP and remove the "Manage" and "Core" dependencies from Mock-Firebolt and covers the below points -

The "manage" dependency is removed from MF as well as anything hard-coded to work with it
The "core" dependency is moved from a dependency to a configuration referring to the new online source for the mono-repo.
The default configuration is updated to load the mono-repo OpenRPC from the online source and enable 'allowMixedCase' flag to true
The MF code is updated to handle self-referencing OpenRPC documents
The "discovery" dependency is not altered


kschrief (approved):

README.md
```diff
@@ -89,11 +89,11 @@ The next two sections presume you are using the out-of-the-box `.mf.config.json`
 
 Mock Firebolt supports these Firebolt SDKs: **core**, **manage**, **discovery**.
 
-By default, only the core SDK is enabled, meaning if you try to make calls to methods in the other SDKs, Mock Firebolt will thow an error.
+By default, only the core/manage OpenRPC retrieved from HTTP is enabled
 
 This default mode is appropriate for app developers developing third-party content apps that don't need (nor get) the permissions necessary to use the other SDKs.
```
michaelcfine, line 94: This statement really isn't true any more. Since the one OpenRPC file has "everything", Mock Firebolt will respond to manage SDK calls, which generally third-party content parties will use. So I think I would just take this paragraph out.

README.md
```diff
@@ -89,11 +89,11 @@ The next two sections presume you are using the out-of-the-box `.mf.config.json`
...
 
-For developers building "operator apps" / "search and discover apps" which need one or more of these SDKs, when running the Mock Firebolt server, you can pass flags like `--manage` and/or `--discovery` to enable the specific SDK(s) desired. Note of course that ultimately, when running on a real device, your app will only have whatever permissions it has been given.
+For developers building "operator apps" / "search and discover apps" which need one or more of these SDKs, when running the Mock Firebolt server, you can pass flags like `--discovery` to enable the specific SDK desired. Note of course that ultimately, when running on a real device, your app will only have whatever permissions it has been given.
```
michaelcfine, line 96: aren't we removing any/all separations (including discovery)?

README.md
```diff
@@ -105,19 +105,6 @@ Developers wishing to activate this functionality must pass the `--moneybadger`
 
 # Important Dependency/Version Notes
```
michaelcfine, line 106: If we're removing the section, we should remove the heading too

README.md
```diff
@@ -136,28 +123,17 @@ cd server
...
-
-npm run build:mf          # Builds source code within this project
-
-# Ongoing stuff
-
-# If this is not your first time using mock-firebolt but you want to ensure you have the latest SDK files you can use
-npm run upgradeSDK
-
-# If you're only using the Firebolt Core SDK (typical for most 3rd party app developers)
+# If you want support for Firebolt Core/Manage SDK
```
michaelcfine, line 129: Do we need/want this heading?

README.md
```diff
@@ -136,28 +123,17 @@ cd server
...
 # Run in a separate terminal window/tab, or use '&' to background
 npm run dev
 
-# If you want support for the Firebolt Manage and/or Firebolt Discovery SDKs
+# If you want support for the Firebolt Discovery SDK
 # Include the flag(s) you require
 # Note the extra double dash!
-npm run dev -- --manage --discovery
+npm run dev -- --discovery
```
michaelcfine, line 136: Shouldn't discovery be handled like all other SDKs? If it's in the single open rpc file, it'll be supported automatically / it won't be separate.

README.md
```diff
@@ -215,13 +189,13 @@ docker run -d \
   -p 9998:9998 \
   $MF_DOCKER_USER/mock-firebolt
 
-# Run the image, enabling the manage and discovery SDKs (in addition to the core SDK) (not typical)
+# Run the image, enabling the discovery SDK (not typical)
```
michaelcfine, line 192: Same question as above... if/when discovery is a thing, it'll be in the one and only one OpenRPC file, right?

docs/DotConfig.md
```diff
@@ -2,20 +2,20 @@ Mock Firebolt: mf.config.json(Supported SDK's)
 =======================
 ## Supported SDK's
 
-Mock Firebolt supports these Firebolt SDKs: core, manage, discovery.
-mf.config.json file contains a list of SDK's supported by Mock Firebolt, Out of which `core` is mandatory and other two `manage` and `discovery` are optional.
+Mock Firebolt supports these Firebolt SDKs: **core** (always),  **manage** (always), **discovery** (optional)
```
michaelcfine, line 5: I think we should remove 'always' and 'optional' since everything comes with the inclusion of the one open rpc file.

docs/DotConfig.md
```diff
@@ -2,20 +2,20 @@ Mock Firebolt: mf.config.json(Supported SDK's)
...
-# If you're only using the Firebolt Core SDK (typical for most 3rd party app developers)
+# If you're only using the Firebolt Core/Manage OpenRPC
 Run Mock Firebolt server as:
 ```
 npm run dev
 ```
 
-# If you want support for the Firebolt Manage and/or Firebolt Discovery SDKs
-# Include the flag(s) you require
+# If you want support for the Firebolt Discovery SDK
```
michaelcfine, line 14: Same comment as above... I think we should not treat discovery separately.

server/package.json
```diff
@@ -6,21 +6,20 @@
   "scripts": {
     "clean": "node ./scripts/clean.mjs",
     "prepare:setup": "npm run clean",
-    "build:core": "node ./scripts/copy.mjs ./node_modules/@firebolt-js/sdk/dist/firebolt-open-rpc.json ./build/firebolt-open-rpc.json",
-    "build:manage": "node ./scripts/copy.mjs ./node_modules/@firebolt-js/manage-sdk/dist/firebolt-manage-open-rpc.json ./build/firebolt-manage-open-rpc.json",
     "build:discovery": "node ./scripts/copy.mjs ./node_modules/@firebolt-js/discovery-sdk/dist/firebolt-discovery-open-rpc.json ./build/firebolt-discovery-open-rpc.json",
     "build:openrpc:NOTES": "Only use if you are using all three Firebolt SDKs",
-    "build:openrpc": "npm run build:core && npm run build:manage && npm run build:discovery",
+    "build:openrpc": "npm run build:discovery",
```
michaelcfine, line 11: need/want to be separate?

server/src/fireboltOpenRpc.mjs
```diff
@@ -301,10 +301,15 @@ async function readSdkJsonFileIfEnabled(sdkName) {
 // NOTE: Assumes the build process has put the firebolt-xxx-sdk.json files in the same directory
 //       as the source code from the src/ directory
 async function readAllEnabledSdkJsonFiles() {
-  await Promise.all(config.dotConfig.supportedSdks.map(async (oSdk) => {
-    const sdkName = oSdk.name;
-    await readSdkJsonFileIfEnabled(sdkName);
-  }));
+  if (isSdkEnabled('mock')) {
```
michaelcfine, line 304: Why is this treated separately/specially?

michaelcfine (changes requested):
Looks better (sdk things), but I still am not sure if discovery should be treated separately or not. Won't it ultimately go in the same openrpc file? If so, we should get rid of any/all places where there's special handling / command line args / etc.

kschrief (approved):

PR #134, opened 2023-06-27
SYSTEST-9456-Change supportedSDKs config and compare in-use/sample co…

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/SYSTEST-9456

1. Compare in-use/sample configs by accessing creation/modification date through metadata of files.
 If mf.config.SAMPLE.json is the same age/older than .mf.config.json: Do nothing. We assume the config is valid
 Else : Print a warning to the screen suggesting they look at release notes to check for new/modified configs.
 
2. Rename "supportedSDKs" in the sample config to "supportedOpenRPCs" and update all references to the same

### DIFF
--- a/server/src/dotConfig.mjs
+++ b/server/src/dotConfig.mjs
@@ -59,14 +59,14 @@ function loadDotConfig() {
 }
 
 /* 
-* @function:findCreatedAndModifiedDate
+* @function:findFileCreationAndModificationTime
 * @Description: To get creation and modification time of files
 @param {String} creationDateFileName - Name of file whose creation time needs to be retrieved in seconds
 @param {String} modificationDateFileName - Name of file whose modification time needs to be retrieved in seconds
 * @Return: Array containing creation and modification time in seconds ex: [1687860231, 1687860231]
 */
 
-function findCreatedAndModifiedDate(creationDateFileName, modificationDateFileName) {
+function findFileCreationAndModificationTime(creationDateFileName, modificationDateFileName) {
   let cFile, mFile, __dirname, __filename
   __filename = fileURLToPath(import.meta.url).replace("build", "src");
   __dirname = path.dirname(__filename);
@@ -77,8 +77,8 @@ function findCreatedAndModifiedDate(creationDateFileName, modificationDateFileNa
   return [creationTimeSec, modificationTimeSec]
 }
 
-const [creationTimeSec, modificationTimeSec] = findCreatedAndModifiedDate('.mf.config.SAMPLE.json', '.mf.config.json')
-if (creationTimeSec > modificationTimeSec) {
+const [creationTimeSec, modificationTimeSec] = findFileCreationAndModificationTime('.mf.config.SAMPLE.json', '.mf.config.json')
+if (creationTimeSec >= modificationTimeSec) {
   logger.importantWarning(`Refer release notes to check for new/modified configs. You probably want to "cp src/.mf.config.SAMPLE.json src/.mf.config.json && npm run build:mf"`)
 }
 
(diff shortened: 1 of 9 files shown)


server/src/dotConfig.mjs
```diff
@@ -58,6 +58,30 @@ function loadDotConfig() {
...
 
+/* 
+* @function:findFileCreationAndModificationTime
+* @Description: To get creation and modification time of files
+@param {String} creationDateFileName - Name of file whose creation time needs to be retrieved in seconds
+@param {String} modificationDateFileName - Name of file whose modification time needs to be retrieved in seconds
+* @Return: Array containing creation and modification time in seconds ex: [1687860231, 1687860231]
+*/
+
+function findFileCreationAndModificationTime(creationDateFileName, modificationDateFileName) {
```
michaelcfine, line 69: What you've written is fine and presumably works, but doing both in one function feels weird and makes the function less 'atomic.' I suggest splitting this into two separate functions, getCreationDate(fileName) and getModificationDate(fileName). Not a super high priority, but favoring smaller, focused functions is usually a good idea.

server/src/dotConfig.mjs
```diff
@@ -58,6 +58,30 @@ function loadDotConfig() {
...
+  cFile = path.resolve(__dirname, creationDateFileName);
+  mFile = path.resolve(__dirname, modificationDateFileName);
+  const creationTimeSec = Math.floor(fs.statSync(cFile).birthtimeMs / 1000);
+  const modificationTimeSec = Math.floor(fs.statSync(mFile).mtimeMs / 1000);
+  return [creationTimeSec, modificationTimeSec]
+}
+
+const [creationTimeSec, modificationTimeSec] = findFileCreationAndModificationTime('.mf.config.SAMPLE.json', '.mf.config.json')
+if (creationTimeSec >= modificationTimeSec) {
+  logger.importantWarning(`Refer release notes to check for new/modified configs. You probably want to "cp src/.mf.config.SAMPLE.json src/.mf.config.json && npm run build:mf"`)
```
michaelcfine, line 82: Small grammar fix to the message:

The .mf.config.SAMLE.json file in your repo is newer than your .mf.config.json file. Changes to the config file format may have occurred since you created or edited your file, and you may need to merge your changes within the changes to the sample file. Refer to the release notes for more information.

Note: I would _not_ include the "You probably want to ..." because if they've edited their file, this would nuke it.

michaelcfine (changes requested):
A few minor changes requested.

Also: We had discussed offering backward compatibility and supporting "supportedSDKs" in addition to "supportedOpenRPCs" for a period of time. Did we decide against that? (I don't see any mention of "supportedSDKs" in the new/changed files. No worries if we decided not to support; but if we did, it seems like the current/new code wouldn't support the old syntax.

server/src/dotConfig.mjs
kschrief: Seconded

kschrief (changes requested):

server/src/dotConfig.mjs
Nandana-NNR: Addressed

server/src/dotConfig.mjs
Nandana-NNR: Addressed

michaelcfine (approved):
Looks good. FYI, there's a server/src/util.mjs file, which is probably a better place for utility functions like the ones you've got here (createAbsoluteFilePath, getCreationDate, getModificationDate). So... in the future, that's where I recommend you put small generic utilities like this. Up to you if you want to move these there and re-submit this PR or leave as-is for now.

Nandana-NNR (commented):
Moved the functions to utils file as suggested. @michaelcfine  Pls re-approve as the earlier approval got revoked

kschrief (approved):

michaelcfine (commented):
👍

PR #136, opened 2023-08-03
Feature/SYSTEST-9487 add live session logging

ksentak:
MF allows users to record "sessions" of communication to/from MF to a log file. This file is only updated when a session is stopped. I added functionality to output data to the file as soon as it's received by MF.


cli/src/usage.mjs
```diff
@@ -42,8 +42,8 @@ const lines = [
...
-  { cmdInfo: "--sessionOutput log|raw|mock-overrides  ",                                  comment: "Set the output format to; log: (paired time sequence of calls, responses)|raw: similiar to log but not paired with request|mock-overrides: a directory of mock overrides" },
-  { cmdInfo: "--sessionOutputPath ../examples/path  ",                                    comment: "Specifiy the session output path. Default for 'log' format will be ./output/sessions and ./output/mocks/<START_TIME> for 'mock-overrides'." },
+  { cmdInfo: "--sessionOutput log|raw|mock-overrides|live  ",                                  comment: "Set the output format to; log: (paired time sequence of calls, responses)|raw: similiar to log but not paired with request|mock-overrides: a directory of mock overrides|live: log messages as they are received in real time" },
+  { cmdInfo: "--sessionOutputPath ../examples/path  ",                                    comment: "Specifiy the session output path. Default for 'log' format will be ./output/sessions and ./output/mocks/<START_TIME> for 'mock-overrides'. Can also be a websocket url" },
```
kschrief, line 46: "Can also be a websocket url" for live only

server/src/sessionManagement.mjs
```diff
@@ -23,6 +23,58 @@
 import { logger } from './logger.mjs';
 import fs from 'fs';
 import yaml from 'js-yaml';
+import WebSocket from 'ws';
+
+class SessionWebSocket {
```
kschrief, line 28: Websocket and File stream classes are very similar. Is it possible to write one class with common functions (open, write, close) and do the "If WS else File" logic here? It would make the usage much cleaner

cli/README.md
```diff
@@ -46,14 +46,15 @@ To change the output format of the session recording use option
...
+- mock-overrides: a directory of method calls extrapolated from the raw format. This option converts the raw formats into json files or yaml files. A json file would be the response from a method call that took place when mock-firebolt processed it. Yaml files contain a function that returns a json response depending on input params. Yaml files are only generated if there were multiple of the same method call with different params.
+- live: This format operates similarly to the 'log' format with an added real-time feature. As each message is received, it gets immediately written to the specified output file. In addition to accepting regular file paths, the 'live' option also supports WebSocket (WS/WSS) URLs. If a WS/WSS URL is designated as the outputPath, a WebSocket connection is established with the specified URL, and the new messages are dispatched to that connection. Please note that specifying an outputPath is essential for the 'live' option. This path is necessary whether you're sending the live log to a WebSocket URL or saving a live copy of the log file to a local directory.
```
kschrief, line 53: To confirm it looks like rather than imitate "raw" we're doing both the logic of "log" and the "live output" logic?

When we "stop()" a session do we get two files? The live output and the "log" output?

kschrief (changes requested):

cli/README.md
ksentak: Question 1: Sorry I have provided incorrect information in the readme. The "live" functionality does indeed imitate "raw" and not "log". I will correct the documentation.

Question 2: When we call stop() we will indeed have two files. One "live" file and another "log" file. Do we want to modify the logic to only provide the "live" file? It is a simple, one-line code change to do so.

server/src/sessionManagement.mjs
ksentak: Change made to include ws and filestream inside of one class

cli/README.md
ksentak: Updated documentation

cli/src/usage.mjs
ksentak: Updated

cli/README.md
```diff
@@ -46,14 +46,15 @@ To change the output format of the session recording use option
...
+- mock-overrides: a directory of method calls extrapolated from the raw format. This option converts the raw formats into json files or yaml files. A json file would be the response from a method call that took place when mock-firebolt processed it. Yaml files contain a function that returns a json response depending on input params. Yaml files are only generated if there were multiple of the same method call with different params.
+- live: This format operates similarly to the 'raw' format with an added real-time feature. As each message is received, it gets immediately written to the specified output file. In addition to accepting regular file paths, the 'live' option also supports WebSocket (WS/WSS) URLs. If a WS/WSS URL is designated as the outputPath, a WebSocket connection is established with the specified URL, and the new messages are dispatched to that connection. Please note that specifying an outputPath is essential for the 'live' option. This path is necessary whether you're sending the live log to a WebSocket URL or saving a live copy of the log file to a local directory.
```
michaelcfine, line 53: I thought the idea was to simply make raw processing 'live', always, and not make this a visible thing. It feels like we're mixing apples and oranges here. Output format to me means things like, are the calls logged separately from responses or are they paired up. But I think of 'live' as simply 'when' things get written... as they happen or when the session recording is over.

cli/README.md
```diff
@@ -46,14 +46,15 @@ To change the output format of the session recording use option
...
-`--sessionOutputPath ./output/examples`
-with start, stop, or call in between starting and stopping.
-This will save the recording to the directory specified. Default for log|raw is server/output/sessions and for mock-overrides is server/output/mocks
+To change the output directory of the session recording, use the --sessionOutputPath option with start, stop, or call. This can be done at any time between starting and stopping the recording. For example, --sessionOutputPath ./output/examples will save the recording to the directory specified. The default paths for log and raw formats are ./output/sessions, and for mock-overrides is ./output/mocks.
+
+Please note that because of the live nature of the "live" option, there should be no expectation of valid JSON formatting. Each message is written as it is received, with a newline character after each message for file appending. For WebSocket appending, we simply write() each message as we receive it. Also, responses are printed to a file before they can be received for the messages, so we can't print responses with their original requests. It will be up to the user reading the file to match request IDs with response IDs.
```
michaelcfine, line 57: Oh, I see. I guess this is a difference. Why don't we say that we write out legal JSON *objects* each time we see a method call or a method response to report... the only thing we're not including is an overall array or object to hold these things.

server/src/sessionManagement.mjs
```diff
@@ -59,6 +116,10 @@ class Session {
             // const sessionStartString = sessionStart.toISOString().replace(/T/, '_').replace(/\..+/, '');
             // logger.info(`${sessionStart.toISOString()}`);    
 
+            // Check if the output path is a WebSocket URL
+            const wsRegex = /^(ws(s)?):\/\//i;
```
michaelcfine, line 120: Minor, but perhaps we can centralize this regex in a constant variable near the top of the module to keep things a tad DRYer?

michaelcfine (approved):
I'm still a bit surprised to see 'live' as an output choice (along with raw), but I guess I get it given the lack of outer array or object. Otherwise, I think very minor comments. 👍

server/src/sessionManagement.mjs
ksentak: Agreed and moved constant to top of the file

cli/README.md
ksentak: Reworded documentation:

Please be aware that while utilizing the "live" option, the system doesn't create a singular encompassing JSON array or object. Instead, it produces individual legal JSON objects for each method call or method response that's observed. Owing to the real-time nature of this process, each message is directly written as it arrives.

For file storage, a newline character is appended after every message to facilitate sequential additions. When it comes to WebSocket, the system simply employs the write() function for every incoming message.

A point to consider is that responses might be written to a file before the corresponding messages are received. Consequently, the file will not correlate the responses with their initiating requests. The responsibility lies with the user to associate request IDs with the relevant response IDs when analyzing the output.

michaelcfine (approved):
👍 LGTM!

kschrief (approved):

PR #139, opened 2023-08-09
Feature - Add multiple user support to sessions

ksentak:
Sessions in MF were implemented before full multi-user support was added. For this reason an instance of MF can only have one "session". All users must share the same session without knowing what messages came for which user.

This PR adds proper user support for Sessions so each user has their own session.


kschrief (approved):

michaelcfine (approved):
I think this looks right, but the diff is fairly large so I'm not sure I've fully, deeply reviewed tbh. As long as all tests pass and the new functionality works, I'm good :).

PR #140, opened 2023-08-17
SYSTEST-9447-Create config associations

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/SYSTEST-9447 (Implementation to create config associations)
Below changes are covered as part of the PR -

-  A new file "changedConfigs.json" was created that contains an array of 'changed' and 'new' configs wrt to base config

- Function 'compareConfigs' was created to compare changedConfig and loaded base config files, generate required warnings if needed and perform dynamic config updates

- Two utils functions 'searchJSONforParam' and 'replaceKeyInJSON' was added additionally to recursively search for required key in config json and update config value dynamically via recursive search of config json respectively to keep the code atomic


server/src/config.mjs
```diff
@@ -53,6 +56,33 @@ const config = {
...
+* @Description: To compare changed and original config files 
+* @param {Object} changedConfig - config that contains updates/additions wrt original config objects
+* @param {Object} loadedConfig - original config
+*/
+function compareConfigs(changedConfig, loadedConfig) {
+  // Iterate through new config array in changedConfig
+  for (const newObj of changedConfig.new) {
+    const newNameToCheck = newObj.name;
+    if (newNameToCheck && !searchJSONforParam(loadedConfig, newNameToCheck)) {
+      logger.warn(`WARNING: There is an unused config ${newObj.name}. Refer ${newObj.readme} for more info `);
```
kschrief, line 70: We need to check if `newObj.readme` is present or not. If it's null/undefined the warning log will be improper.

server/src/config.mjs
```diff
@@ -53,6 +56,33 @@ const config = {
...
+
+    }
+  }
+  // Iterate through changed config array in changedConfig
+  for (const changedObj of changedConfig.changed) {
+    const oldNameToCheck = changedObj.oldName;
+    if (oldNameToCheck && searchJSONforParam(loadedConfig, oldNameToCheck)) {
+      logger.warn(`WARNING: The config ${changedObj.oldName} has been renamed to ${changedObj.newName}. Refer ${changedObj.readme} for more info `);
```
kschrief, line 78: We need to check if `newObj.readme` is present or not. If it's null/undefined the warning log will be improper.

server/src/changedConfigs.json
```diff
@@ -0,0 +1,15 @@
+{
```
kschrief, line 1: Can we make a change to this based off of Mike's comments?

Let's change to the below format:

`[
  {
    "type": "new",
    "name": "xyz",
    "readme": "readme.md"
  },
  {
    "type": "changed",
    "oldName": "abc",
    "newName": "xyz",
    "readme": "readme.md"
  }
]`

Sorry for the last-minute ask.

kschrief (changes requested):

kschrief (approved):

server/src/config.mjs
```diff
@@ -53,6 +56,45 @@ const config = {
...
+* @param {Object} changedConfig - config that contains updates/additions wrt original config objects
+* @param {Object} loadedConfig - original config
+*/
+function compareConfigs(changedConfig, loadedConfig) {
+  // Iterate through new and changed config objects in changedConfig
+  for (const obj of changedConfig) {
+    if (obj.type == 'new') {
+      const newNameToCheck = obj.name;
+      const readme = obj.readme;
+      if (newNameToCheck && !searchJSONforParam(loadedConfig, newNameToCheck)) {
```
michaelcfine, line 71: JSON is a string serialization format... so technically, I think you're really 'just' searching an object for a param. I don't think it's critical, but you could/eventually should rename this function something like 'searchObjectForParam' or perhaps better 'searchObjectForKey'. Minor, but sometimes naming subtleties can help/hurt.

michaelcfine (approved):
Looks good. A few minor comments.

kschrief (approved):

PR #141, opened 2023-09-06
Feature - Set up ws server for session output

ksentak:
### Background
Giving a WS as an output for "sessionOutput" is working well. For ease-of-use with other applications that can't easily spin up a server we should also accept incoming WS connections and use those to send out the same data that we do for outgoing WS connections.

### What this PR accomplishes
* Introduces a new sessionOutput option, named server. This option enables the MF Session WS Server to send messages to connected WebSocket clients.

* Adds a new port (default: 9999) on startup, which functions as a WebSocket server, emitting recorded session activity.

* Enables the new WebSocket server to accept incoming connections.
     * Without a specified userId, the server will broadcast all user session activities to connected clients.
     * With a specified userId, the server will send only the session activities related to that user to the connected client.


jbigel (changes requested):

jbigel (approved):

michaelcfine (approved):
LGTM 👍

Issue #145, opened 2023-10-23
Triggering lifecycle events with CLI and browser extension does not match casing


mkatenolan:
I have been trying to use the CLI and browser extension to trigger lifecycle events in an app with listeners set up using ```Lifecycle.listen``` as described in SDK docs. 

I was making use of the example events in the CLI folder and kept getting the `NO-EVENT-HANDLER-REGISTERED` error code even though the server was registering listeners with messages like:
```Registered event listener mapping: 12345:lifecycle.onbackground```

If I change the example event `method` (eg. [here](https://github.com/rdkcentral/mock-firebolt/blob/main/cli/examples/lifecycle-background-to-foreground.event.json)) from camelcase to lowercase, the event get triggered correctly. Eg.

```
{
  "method": "lifecycle.onForeground", // Does NOT trigger event
  "result": {
    "state": "foreground",
    "previous": "background"
  }
}
```

```
{
  "method": "lifecycle.onforeground", // Correctly triggers event
  "result": {
    "state": "foreground",
    "previous": "background"
  }
}
```

Similarly in the browser extension: [eg.](https://github.com/rdkcentral/mock-firebolt/blob/main/browser-extensions/governor/background.js#L65)

```
async function sendFireboltLifecycleEvent_Inactive(cb) {
  return await _sendFireboltLifecycleEvent('lifecycle.onInactive', { state: 'inactive' }, cb); // Does NOT trigger event
}
```

```
async function sendFireboltLifecycleEvent_Inactive(cb) {
  return await _sendFireboltLifecycleEvent('lifecycle.oninactive', { state: 'inactive' }, cb); // Correctly triggers event
}
```

It seems that either the examples are incorrect or the server is not parsing the events correctly. It would be ideal if we could use the examples out the box rather than having to update them.

kschrief:
Hi Martha.

I did an investigation. The bug is valid and the PR properly addresses the issue. Thanks for bringing this up! I'm going to approve the PR but we won't be able to merge it in unless you sign the Contributor Agreement. Once that's done everything else should be good to go.

PR #146, opened 2023-11-01
add check for allowLowerCase on sendEvent

mkatenolan:
When triggering lifecycle events via the CLI using the examples eg.
```
{
  "method": "lifecycle.onForeground", // Does NOT trigger event
  "result": {
    "state": "foreground",
    "previous": "background"
  }
}
```
Events are not registered correctly due to a mis-match in casing. More details in this issue: https://github.com/rdkcentral/mock-firebolt/issues/145

This PR checks the `allowMixedCase` config value and applies `toLowerCase` on the lifecycle `method` when set to true. This is the same pattern used elsewhere in the code.


kschrief (approved):

kschrief:
I've approved the PR but this cannot be merged in until the CLA is signed. Thanks in advance!

PR #148, opened 2023-11-30
SYSTEST-9607- Event error message bug fix

Nandana-NNR:
Bug fix for https://ccp.sys.comcast.net/browse/SYSTEST-9607
### Added

* New unit test to validate updated error handling function and functional test update to capture updated error message

### Changed

* Refined `logErr()` and `fErr()` functions which handles event error messages to appropriately throw both event validation and event registration errors. 

### Fixed

* Fixed bug related to event error message handling wherein all event errors were previously thrown as an improper generic error which was misleading to users. Implemented a mechanism to correctly distinguish different types of event errors and throw them appropriately

### DIFF
--- a/server/src/events.mjs
+++ b/server/src/events.mjs
@@ -309,11 +309,11 @@ function coreSendEvent(isBroadcast, ws, userId, method, result, msg, fSuccess, f
     if (  ! isBroadcast && !isRegisteredEventListener(userId, method) ) {
       logger.info(`${method} event not registered`);
      
-      fErr.call(null, method, 'registrationError');
+      fErr.call(null, 'registrationError', method);
 
     } else if ( isBroadcast && !isAnyRegisteredInGroup(userId, method) ){
       logger.info(`${method} event not registered`);
-      fErr.call(null, method, 'registrationError');
+      fErr.call(null, 'registrationError', method);
 
     } else {
        // Fire pre trigger if there is one for this method
@@ -383,7 +383,7 @@ function coreSendEvent(isBroadcast, ws, userId, method, result, msg, fSuccess, f
       if( config.validate.includes("events") ) {
         const resultErrors = fireboltOpenRpc.validateMethodResult(finalResult, method);
         if ( resultErrors && resultErrors.length > 0 ) {
-          fErr.call(null, method, 'validationError');
+          fErr.call(null, 'validationError', method);
           return
         }
       }
--- a/server/src/events.mjs
+++ b/server/src/events.mjs
@@ -31,6 +31,7 @@ import { config } from './config.mjs';
 import { updateCallWithResponse } from './sessionManagement.mjs';
 
 const { dotConfig: { eventConfig } } = config;
+let eventErrorType;
 
 function logSuccess(onMethod, result, msg) {
   logger.info(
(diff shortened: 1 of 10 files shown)


server/src/events.mjs
```diff
@@ -301,10 +303,12 @@ function coreSendEvent(isBroadcast, ws, userId, method, result, msg, fSuccess, f
   try {
     if (  ! isBroadcast && !isRegisteredEventListener(userId, method) ) {
       logger.info(`${method} event not registered`);
+      eventErrorType = 'registrationError'
       fErr.call(null, method);
```
ksentak, line 307: Im not 100% sure where the .call() function comes from, but instead of exporting eventErrorType from events.mjs, would we be able to pass it into the call below? So it would look something like this

fErr.call(null, method, errorType);

If we are able to do this it would simplify the logic inside of the api file

server/src/routes/api/event.mjs
```diff
@@ -39,10 +39,13 @@ function sendEvent(req, res) {
...
 
   function fErr(method) {
+    const errorCode = (events.eventErrorType === 'validationError') ? 'EVENT-VALIDATION-FAILED' : 'NO-EVENT-HANDLER-REGISTERED';
     res.status(400).send({
       status: 'ERROR',
-      errorCode: 'NO-EVENT-HANDLER-REGISTERED',
-      message: `Could not send ${method} event because no listener is active`
+      errorCode: errorCode,
+      message: (errorCode === 'EVENT-VALIDATION-FAILED') ?
+        `Event validation failed for ${method}. Please ensure the event data meets the required format and try again.` :
```
michaelcfine, line 47: Minor/stylistic: I vote for lifting this ternary operator up out of the 'anonymous' object creation... having a new "const message = xxx ? yyy : zzz line and then just using message: message here (like you've done with errorCode)

server/src/routes/api/event.mjs
```diff
@@ -72,10 +75,13 @@ function sendBroadcastEvent(req, res) {
...
```
michaelcfine, line 83: same comment as above to separate this out

server/src/events.mjs
```diff
@@ -39,11 +40,12 @@ function logSuccess(onMethod, result, msg) {
 };
 
 function logErr(onMethod) {
-  logger.info(
-    `Could not send ${onMethod} event because no listener is active`
-  );
-};
-
+  if (eventErrorType == 'validationError') {
```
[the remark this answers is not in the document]
Nandana-NNR: updated this code to use switch-case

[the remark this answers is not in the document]
Nandana-NNR: Addressed

server/src/events.mjs
```diff
@@ -301,10 +303,12 @@ function coreSendEvent(isBroadcast, ws, userId, method, result, msg, fSuccess, f
   try {
     if (  ! isBroadcast && !isRegisteredEventListener(userId, method) ) {
       logger.info(`${method} event not registered`);
+      eventErrorType = 'registrationError'
       fErr.call(null, method);
 
     } else if ( isBroadcast && !isAnyRegisteredInGroup(userId, method) ){
       logger.info(`${method} event not registered`);
+      eventErrorType = 'registrationError'
```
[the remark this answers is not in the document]
Nandana-NNR: Addressed for fErr. Also fFatalErr catches the final exception from the whole sendEvent function, unlike fErr which throws custom error messages based on different types of error. So I dont think we need to add parameter to pass errortype for fFatalErr
```
 } catch ( ex ) {
    logger.error('sendEvent: ERROR:');
    logger.error(ex);
    fFatalErr.call(null, ex);
  }
}
```

server/src/routes/api/event.mjs
Nandana-NNR: Addressed

server/src/routes/api/event.mjs
Nandana-NNR: Addressed

ksentak (approved):

kschrief (approved):

PR #150, opened 2024-02-13
FIRECERT-1846-POC for hot/cold launch scenarios

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/FIRECERT-1846
* Added ability to check if an active userid is connected or not to mock-firebolt from yaml file
* Minor bug fix - Added cross-env package to set environment variables across platforms, to resolve platform specific bug while executing npm run test command in MF
*  Added ability to remove websocket connection object associated with a user when close event is received via websocket


Nandana-NNR (commented):
As per current usecase for discovery launch scenarios, only the ctx object inside 
stateManagement->handleDynamicResponseValues needs to be updated as this is the only updated function that gets triggered when override is set. But to maintain consistency and considering the possibility of future usecases, I have updated the ctx objects defined across all other files to include the new function. Let me know if that's ok or do we need to cover only ctx object in handleDynamicResponseValues for now
cc @kschrief

server/src/userManagement.mjs
```diff
@@ -94,9 +94,11 @@ function getWssForUser(userId) {
 //get the ws for user,if user has ws already,it will return the latest ws connection
 function getWsForUser(userId) {
   if ( user2ws.has(''+userId) ) {
-    let wsArray=user2ws.get(''+userId);
-    let latestWsConnection=wsArray[(wsArray.length-1)]
-    return latestWsConnection;
+    let wsArray = user2ws.get("" + userId);
+    if (wsArray) {
```
kschrief, line 98: Let's check for a non-empty array as well.

server/src/userManagement.mjs
```diff
@@ -197,6 +199,9 @@ function addUser(userId) {
     ws.on('pong', async hb => {
       heartbeat(ws)
     });
+    ws.on('close', function close() {
+      user2ws.delete(''+userId);
```
kschrief, line 203: I think we need to do this a little differently. This line deletes the entire Array of WS objects for the user. If we run this when the user has more than one connection the others will be put in a weird state.

We should go through the array and delete _only_ the current ws object from the array.

kschrief (changes requested):

PR #156, opened 2024-05-21
Added ability for ctx.delay

kdivya153:


server/src/stateManagement.mjs
```diff
@@ -274,7 +275,7 @@ function handleDynamicResponseValues(userId, methodName, params, ws, resp){
       };
       const sFcnBody = resp.response + ';' + 'return f(ctx, params);'
       const fcn = new Function('ctx', 'params', sFcnBody);
-      const result = fcn(ctx, params);
+      const result = await fcn(ctx, params);
```
kschrief, line 278: We call dynamic functions (YAML) in multiple places. They should all be updated to support async/await

kschrief (changes requested):

server/src/stateManagement.mjs
kdivya153: Added await/asyn in the other places,where dynamic function is being called

michaelcfine (approved):
Assuming you found/changed all the fcn() calls and stateManagement.getMethodResponse() calls and handleStaticAndDynamicResult() calls, looks good to me.

PR #167, opened 2025-03-25
feat: toApp support added

AdityaKasar:


server/src/events.mjs
```diff
@@ -302,22 +352,27 @@ function emitResponse(finalResult, msg, userId, method) {
...
+  // Check if bidirectional mode is enabled
+  if (config.dotConfig.bidirectional) {
+    const bidirectionalMethod = unidirectionalEventToBiDirectional(method);
+    let payload = createBidirectionalPayload(bidirectionalMethod, finalResult);
+
+    wsArr.forEach((ws) => {
+      ws.send(JSON.stringify(payload)); // Send bidirectional event
+      logger.info(`${msg}: Sent bidirectional event to user ${userId}: ${JSON.stringify(payload)}`);
+    });
+  } else {
```
kschrief, line 365: We need to keep the "updateCallWithResponse()" call in this block.

server/src/fireboltOpenRpc.mjs
```diff
@@ -84,9 +89,15 @@ function isMethodKnown(methodName) {
 }
 
 function getSchema(schemaName) {
-  for ( const ii = 0; ii < config.dotConfig.supportedOpenRPCs.length; ii += 1 ) {
-    const sdkName = config.dotConfig.supportedOpenRPCs[ii].name;
-    if ( schemaName in meta[sdkName].components.schemas ) { return meta[sdkName].components.schemas[schemaName]; }
+  const sources = [
```
kschrief, line 92: We do this quite often. Can we create a utility function for this?

kschrief (changes requested):

server/src/events.mjs
AdityaKasar: Unidirectional mode (Default behavior) added back in else block

server/src/fireboltOpenRpc.mjs
AdityaKasar: created a new utility function getOpenRPCSources().

kschrief (approved):

